PR #2, opened 2015-04-25
Fix indentation, pep8. Optimize imports and adding missing import for sy...

michael-borisov:
...s module.

Improvements in some parts of  get_all_imports method and removing unused path variable.


bndr:
Thanks!

PR #3, opened 2015-04-26
Fixed problems with local imports and imports with 'as'

littmus:
1. local imports
```
λ ~/temp/ ls
main.py  models.py  models.pyc
```
Current directory's files are like these, and I imported `models` in `main.py`.
Cause a package named `models` exists in pypi, it is added to requirements file.
2. imported with "as"
```python
...
import ujson as json
...
```
In this case, it's package name is recognized invalid by `ujson as json`.


bndr:
Thanks! but Travis build fails, fix it, and I will accept the merge request.

PR #8, opened 2015-04-28
Output tuning

Ohcanep:
Fix a typo, use str.format() in a few places, sort requirements list and update help text.


bndr:
Thanks!

Issue #13, opened 2015-05-08
Problem handling multiple imports in the same line when using aliases


tiagofmcosta:
I just came across a problem when trying to use multiple imports in the same line with defined aliases, like `import requests as R, bs4 as BS, dns`.

As I see it the problem lies in the fact that you're only stripping the aliases from the import name when you don't find a `,` on the line.

tiagofmcosta:
I'll create a pull request in a few minutes to solve this issue.

PR #16, opened 2015-05-08
Handling multiple imports in the same line when using aliases

tiagofmcosta:
* Fix for #13.
* Added tests for the new method `get_import_name_without_alias`


bndr:
Thanks!

PR #21, opened 2015-07-14
Fix flake8

lord63:


lord63:
When I follow the contributing guide and run `flake8 tests pipreqs`, I get many flake8 warnings, so i try to fix them.

Note still has following warnings:

```
tests/_data/test.py:2:80: E501 line too long (117 > 79 characters)
tests/_data/test.py:37:14: E401 multiple imports on one line
tests/_data/test.py:46:17: E401 multiple imports on one line
tests/_data/test.py:46:25: E901 SyntaxError: invalid syntax
```

lord63:
The reason:

* the first: I'm afraid that the modify may affect the pylint
* the second and the third: is it for test purpose?
* the forth: SyntaxError, I can't get the point.

maybe we should just ignore `tests/_data/test.py`? Cause we have many unused import.

bndr:
`tests/_data/test.py` is syntactically wrong to test pipreqs for different cases.

PR #24, opened 2015-08-31
Skip irreverent directory and change logging level

Lee-W:
- Skip .git, env, \__pycache\__ when running get_all_imports()  
  - In the previous version, pipreqs walks through `env` and generate unnecessary imports
- Change logging level to warning when encountering `Requirements.txt already exists`


bndr:
Looks good, Thanks!

Issue #25, opened 2015-09-28
Add other version control systems to the default ignores


GadgetSteve:
Currently pipreqs.py has an ignore list of: *line 35 of pipreqs.p*
`ignore_dirs = [".git", "__pycache__", "env"]`
which is fine if you are running git but users of SVN & Mercurial will either end up with very slow performance or other problems.

Please consider **at least**:
`ignore_dirs = [".hg", ".svn", ".git", "__pycache__", "env"]`

PR #26, opened 2015-09-28
Added .svn & .hg to directory ignores.

GadgetSteve:
Added .svn & .hg to directory ignores.
Also added *.bak to git ignores.

Addresses #25


bndr:
Thanks!

PR #28, opened 2015-10-19
feat(cli) add --encoding option for open()

hdformat:
UnicodeDecodeError can be caused by windows console's the default charset -  my windows 7 korean console's charset is 'cp949(euc-kr)' and my python file is encoded in utf-8. so I've added encoding parameter (optional) to be passed to open() function.
```
UnicodeDecodeError: 'cp949' codec can't decode byte 0x9a in position 1782: illegal multibyte sequence
```


bndr:
Can you fix the tests please? I will merge it after that

Issue #36, opened 2016-01-20
Import parsing issue when imports are not at the top of the file


BartGo:
Windows 7, pipreqs 0.3.9, file to process (example/main.py):
```
import logging
logging.basicConfig()
from pydal import DAL, Field # some comment
from transitions import Machine # some comment

print "Hello"
```

Running:
```
> pipreqs --savepath req-pipreqs.txt --debug example

DEBUG: Found packages: set(['logging'])
DEBUG: Found imports:
DEBUG: Getting packages information from Local/PyPI
DEBUG: Writing 0 requirements:  to req-pipreqs.txt
INFO: Successfully saved requirements file in req-pipreqs.txt

```
Logging is not saved into req-pipreqs.txt which is fine because it's a standard library. Imports of pydal and transitions are not taken into account which is an error.

Removing the second line (or commenting out), so "logging.basicConfig()" causes pydal and transitions to be recognised correctly so it's a source parsing issue.
 
Seems to be related to the fix of #35 but also may be a slightly different scenario.

bndr:
That is correct. PEP8 states that imports should be at the top of the file, so if you start calling functions and defining variables, everything after that is assumed to be code not imports.

BartGo:
Thanks for your quick answer!

I see that you're referring to E402, it's very restrictive (and there are some discussions around it) but anyway the issue I noticed is present not only for the cases violating PEP8/E402, it's wider.

Example:
```
import bottle
try:
    import json
except ImportError:
    import simplejson as json

print "Imports done."
b = bottle.Bottle()
j = json.dumps([])
```
And:
```
> flake8 -v example/main.py
```
No issues found (flake: 2.5.1, pep8: 1.7.0, pyflakes: 1.0.0, mccabe: 0.3.1). Then, pipreqs does not recognise any of the modules (bottle, json and/or simplejson):

```
> pipreqs --debug --savepath req.txt example/main.py
DEBUG: Found packages: set([])
DEBUG: Found imports:
DEBUG: Getting packages information from Local/PyPI
DEBUG: Writing 0 requirements:  to req.txt
INFO: Successfully saved requirements file in req.txt
```

bndr:
You do have a point, another thing I forgot is lazy loading imports e.g:

```
def somefunc():
    import some_module
```

So this should definitely be looked into.

Thanks again for reporting.

BartGo:
No problem! I renamed the issue as the previous title was misleading.

bndr:
[...]

PR #37, opened 2016-01-28
Walk Abstract Syntax Tree to find imports

9cloud:
Rationale: Using regex means that you can only find the imports that are at the top of the page without any spaces between them. There's likely other issues, but that was the one that bit me. 

Caveat: Old method would parse files with broken syntax. This one will not.

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -68,14 +68,14 @@ def get_all_imports(path, encoding=None):
                                 raw_imports.add(subnode.name)
                         elif isinstance(node, ast.ImportFrom):
                             raw_imports.add(node.module)
-                except Exception, e:
+                except Exception as exc:
                     if ignore_errors:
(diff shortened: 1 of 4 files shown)


bndr:
This looks quite good, I tested it locally and it works as expected. You also added comments to tests, which is also nice :)

This also fixes #36

Thanks for the PR!

9cloud:
In case its a concern that AST walking is slow. I've done some spot tests.

_Before AST:_ runtime 0m0.718s. Found 40 requirements.
_After AST:_ runtime 0m3.043s. Found 60 requirements.

**Caveats**

Since this reaches down into the depths of the AST; you will get all imports even conditional ones. This may result in a build that’s impossible to actually have; but this is a problem shared with the previous non-AST system.

More likely this will result in a good thing: those imports you forgot about b/c they’re buried in the depths of sites-packages; you can get them now.

PR #38, opened 2016-02-05
added ignore option

nrhine1:
Added an option to pass directories to ignore. 

The tests are unwritten but shouldn't be difficult to write - it seems there are already some issues with failing tests elsewhere so I didn't include them.


bndr:
Thanks! Looks good

nrhine1:
I actually ran into some issues (first time using docopt), and modified the code. Before it parsed the directories as a list of characters, and thus didn't ignore the directories (and could possibly ignore single character directories!)

The current commit in my fork contains these updates. (the tests seem much more appealing now)
https://github.com/nrhine1/pipreqs/commit/bb1eb59a7b01c6658e9675e86312c9daf54bdcd1

I can open new PR if desired

bndr:
New PR would be nice :) Thanks!

PR #39, opened 2016-02-05
Ignore option uses comma-separated directories

nrhine1:
Docopt was parsing the ignore options incorrectly, and a new argument interface is adopted to pass ignore directories as a comma-separated list. It seems there should be a better way to do this with docopt, but perhaps not (see: 
http://stackoverflow.com/questions/29313995/including-more-than-one-list-of-arguments-with-docopt )

Tests are left unwritten.


bndr:
I took your code and rewrote it a bit, it should be working in release 0.4.1.

Thanks again!

PR #41, opened 2016-02-10
Make pipreqs not write diplicate lines to requirements.txt file

dmand:
Test file:
```python
import pymongo
from bson.objectid import ObjectId
```

For this file `pipreqs` generates the following `requirements.txt`:
```
pymongo==3.2.1
pymongo==3.2.1
```

This pull-request fixes this unfortunate behaviour by patching `get_pkg_names` function.


bndr:
Thanks! Looks good

PR #53, opened 2016-11-21
Implement --print to output list of requirements to stdout

### LINKED ISSUE #43: Add an option to just print the output
Usually you want such a tool to just print, then you redirect it to the file you want. Since we can't do that without breaking compatibilities without previous versions, how about adding a --print option ?

This will allow you to easily pipe the output, grep it, etc. Or just inspect the content before saving it somewhere.

utgwkk:
close #43


bndr:
Looks good to me. Thanks!

PR #84, opened 2017-08-31
Follow symbolic linked directories

rspencer01:
This makes pipreqs dive into directories that are symlinks as Python
recognises these.  If symlinks are not followed, then the requirements
will be incorrect.

We also add a command line option to disable following of symlinks if
need be.


jonafato:
Thanks!

PR #89, opened 2017-10-09
Ordering imports by name

annp89:
Description:
- Order the imports generated by name

Tests:
- Tests support both 2 and 3

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -22,6 +22,7 @@ Options:
     --clean <file>        Clean up requirements.txt by removing modules that are not imported in project.
 """
 from __future__ import print_function, absolute_import
+import operator
 import os
 import sys
 import re
@@ -125,6 +126,7 @@ def generate_requirements_file(path, imports):
                                  for item in imports) + '\n')
 
 def output_requirements(imports):
+    imports.sort(key=operator.itemgetter('name'))
     logging.debug('Writing {num} requirements: {imports} to stdout'.format(
         num=len(imports),
         imports=", ".join([x['name'] for x in imports])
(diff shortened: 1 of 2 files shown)


jonafato:
The imports are already [sorted](https://github.com/bndr/pipreqs/blob/22e80c27c2365b3974b2e787076e9f8f45222ed1/pipreqs/pipreqs.py#L109) when read from the source files. Are you seeing behavior different from this?

annp89:
It wasnt ordered when I ran it on one of my private repositories. How would you like me to share the behavior? Screen shot of me running the command?

jonafato:
@annp89 Can you provide a minimal set of imports that demonstrates this behavior when `pipreqs` is run? This can likely be fixed easily, but I'd like to make sure we're not sorting multiple times redundantly.

annp89:
Partial output of what I see after doing another run 
```
geopy==1.11.0
requests==2.18.4
python_firebase==1.2
pycrypto==2.6.1
boto==2.48.0
boto3==1.4.7
cachetools==2.0.1
ciso8601==1.0.5
cloudinary==1.8.0
python_dateutil==2.6.1
Django==2.0b1
django_dynamic_fixture==1.9.5
Fabric==1.14.0
firebase_token_generator==2.0.1
gevent==1.2.2
grpc==0.3-19
jose==1.0.0
```

jonafato:
I can't seem to reproduce this under Python 2.7, 3.5, or 3.6 by simply putting all of those into a Python file and running `pipreqs .`. Can you provide some more info?

- The Python file(s) (just the import lines and directory structure will do) that cause pipreqs to produce this output
- pipreqs verions
- Python version
- Operating system

Issue #94, opened 2017-10-20
Optionally omit package version from requirements file


jonafato:
Developers don't always want to include pinned dependencies in the output generated by `pipreqs` (e.g. when using it in conjunction with `pip-compile` from https://github.com/jazzband/pip-tools). `pipreqs` should have a `--no-pin` option to omit versions when desired.

troyhoffman:
Pip freeze always stores the version, but I like the idea of being able to omit the versions.

bersbersbers:
It seems this has been implemented already, but I wanted to add another suggestion. I would find it helpful, even when omitting package versions, to indicate which version I have used. So my ideal output would be using `>=` rather than `==`, still using the version number I have installed. This gives you the best of both worlds: You are not forcing any installer to an old version, but you still transmit the information which (old) version worked for you.

alan-barzilay:
Everything suggested here seems to have been implemented already, closing

PR #98, opened 2017-10-23
Implement `--no-pin` option.

dxsh:
See #94.


pipreqs/pipreqs.py
```diff
@@ -115,16 +116,22 @@ def filter_line(l):
...
         logging.debug('Writing {num} requirements: {imports} to {file}'.format(
             num=len(imports),
             file=path,
             imports=", ".join([x['name'] for x in imports])
         ))
-        fmt = '{name}=={version}'
-        out_file.write('\n'.join(fmt.format(**item) if item['version'] else '{name}'.format(**item)
-                                 for item in imports) + '\n')
+
+        if omit_version is True:
```
jonafato: Stylistic note: this can simply be `if omit_version:`, since `True` and `False` are the only possible values here (enforced by `docopt`).

PR #107, opened 2017-11-20
added support for >= in addition to ==

talperetz:
using flag --gt
(gt = greater than)

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -12,6 +12,7 @@ Arguments:
                           directory).
 
 Options:
+    --gt                  Replace format from name==version to name>=version
     --use-local           Use ONLY local package info instead of querying PyPI.
     --pypi-server <url>   Use custom PyPi server.
     --proxy <url>         Use Proxy, parameter will be passed to requests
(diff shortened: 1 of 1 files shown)


answerquest:
pls merge this already!

PyDataBlog:
Please merge this already :)

Alexander-Shukaev:
```
def read(*paths, **kwargs):
  try:
    return io.open(join(kwargs.get('path') or __realdir__, *paths),
                   encoding=(kwargs.get('encoding') or 'utf8')).read()
  except IOError:
    return ''

def parse_requirements(*paths, **kwargs):
  if not paths:
    paths = ('requirements.txt', )
  path           = join(kwargs.get('path') or __realdir__, *paths)
  kwargs['path'] = None
  strings        = read(path, **kwargs).splitlines()
  kwargs['path'] = dirname(path)
  for number, string in enumerate(strings, 1):
    string = string.strip()
    if not string or string.startswith(('-e', '#')):
      continue
    if string.startswith(('-r')):
      for requirement in parse_requirements(string[2:].strip(), **kwargs):
        yield requirement
    elif kwargs.get('loose', not {'develop'}.intersection(sys.argv)):
      yield string.replace('==', '>=')
    else:
      yield string
```

```
install_requires=list(parse_requirements())
```

Mattwmaster58:
I'm not sure why this is taking so long to merge, but I think a better option entirely would be an option to use the 'compatible release' [version specifier](https://www.python.org/dev/peps/pep-0440/#version-specifiers).

mapattacker:
Wow, this has been 3 years and not merged yet? Why? +__+

alan-barzilay:
Hi! I'm the new maintainer for this project (as of a week ago) and I'm slowly going through all the issues and PRs. I would like to merge this PR but I need to have at least one test  to check that this new flag is working properly.

If you would like to see this feature finally merged, you could help by writing the test. My priority right now is expand support for jupyter notebooks and drop python 2, so unless someone steps up and writes the test this is going to take a while before I have time to look into it.

alan-barzilay:
@Mattwmaster58 I like the idea of adding support for compatible release but I don't see why it should replace the --gt option. Are you suggesting having only the ~= option or did I misunderstood you?

Mattwmaster58:
@alan-barzilay I'm sure what my reasoning was two years ago, but as of today, I think the better options would be to have a `--compat` flag or something to do that. I feel like in almost every case you would want `~=`, not `>=`, so this compat flag would be a lot more useful than `--gt` IMO.

alan-barzilay:
Feature implemented in PR#239

PR #144, opened 2018-11-21
Warn when pypi queries fail

andrew-vant:
Currently, pipreqs silently ignores pypi errors unless --debug is
supplied. In particular, when a name needs to be in the mapping file but
isn't, pipreqs will look for it on pypi, fail to find it, and skip it
without comment. requirements.txt still gets built, but is likely to be
wrong.

This PR makes it treat pypi errors as warning-worthy, and adds some
debugging lines around mapping lookups.

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -259,16 +259,26 @@ def get_pkg_names(pkgs):
         List[str]: The corresponding PyPI package names.
 
     """
-    result = set()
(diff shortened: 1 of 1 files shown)


andrew-vant:
I was cleaning up some of my stale PRs and noticed this was still open with no commits to the repo for a year. Is pipreqs not maintained anymore? Has development moved elsewhere?

Issue #127, opened 2018-06-14
pipreqs not capturing all imports


srishti0190:
In my code I wanted to create a requirements.txt file with all the dependencies needed to run my model. 
I used pip freeze earlier, however, because of obvious reasons i.e. it captures everything from the environment and not just the ones needed for my model, I moved to pipreqs. 

I tested it with a very basic .py file with some imports, out of those the following 2 are not getting captured correctly:
`import cv2
import skimage`

The problem I faces is:
1) It fails to capture opencv-python for cv2.
2) for skimage, it captures both scikit_image and skimage. However, I do not need skimage as this is just a dummy package which points to scikit-image. 

Can anyone help here.

Sohaib90:
Same issue, can someone report what is the issue here?

bersbersbers:
Related to `cv2`, `pipreqs` should ideally figure out whether `opencv-contrib-python` is needed (or default to that) instead of installing only `opencv-python`.

gromain:
This issue still exists in the latest version.

FRIKIdelTO:
same issue here

aw632:
Same issue

cjj1120:
same issue as well for version `pipreqs==0.4.11`

OmAximani0:
> same issue as well for version `pipreqs==0.4.11`

Me facing problem too

StantonMatt:
I checked this against current `master`.

The original `cv2` case appears covered in the current mapping file:

```text
cv2:opencv-python
```

Using the same import-name lookup direction as `get_pkg_names`, the current tree maps:

```text
cv2 -> opencv-python
airflow -> apache-airflow
requests -> Requests
```

So the `import cv2` report looks fixed on `master`. If this is still showing up from an installed package version, that may be a release/version-availability issue rather than a missing mapping in the current tree.

Checks:

```text
rg -n "^cv2:|^airflow:|^requests:" pipreqs/mapping
python -c "<mapping probe for cv2/airflow/requests>"
```

PR #180, opened 2019-11-21
#127: add apache-airflow to mapping

brentfollin:
add `apache-airflow` instead of `airflow` to requirements.txt.

See https://pypi.org/project/airflow/

### DIFF
--- a/pipreqs/mapping
+++ b/pipreqs/mapping
@@ -1,6 +1,7 @@
 AG_fft_tools:agpy
 ANSI:pexpect
 Adafruit:Adafruit_Libraries
+apache-airflow:airflow
 App:Zope2
 Asterisk:py_Asterisk
 BB_jekyll_hook:bitbucket_jekyll_hook


pipreqs/mapping
```diff
@@ -1,6 +1,7 @@
 AG_fft_tools:agpy
 ANSI:pexpect
 Adafruit:Adafruit_Libraries
+apache-airflow:airflow
```
alan-barzilay, line 4: I believe you reversed it, it should be import_name:pypi_name, like so:
airflow:apache-airflow

alan-barzilay (changes requested):
Travis seems to have failed the build for some internal connectivity issues. If the mapping is reversed I can merge this PR

StantonMatt:
I rechecked this against current `master` using the GitHub merge ref.

The mapping file is import name -> requirement name. Current `master` already has the useful entry:

```text
airflow:apache-airflow
```

The merge ref for this PR adds the reverse entry as well:

```text
apache-airflow:airflow
```

With the merged mapping:

```text
airflow -> apache-airflow
apache-airflow -> airflow
cv2 -> opencv-python
```

So `import airflow` is already covered on current `master`, while the new reverse entry would only apply to `apache-airflow` as an import name. Since Python imports cannot use the hyphenated package name directly, this still looks like the mapping-direction issue from the earlier review rather than a current merge target.

Checks I ran:

```text
rg -n "^(airflow|apache-airflow|cv2):" pipreqs/mapping
python -c "<mapping probe for airflow/apache-airflow/cv2>"
git diff --check origin/master...HEAD -- pipreqs/mapping
```

GitHub reports no hosted checks for this PR.

Issue #199, opened 2020-04-05
Support for jupyter notebooks


frankhart2018:
I was using pipreqs for a project and I found out that if libraries are imported inside a jupyter notebook, then they aren't being taken into account by pipreqs while creating requirements.txt file. It would be a nice addition to the library.

jonafato:
Agreed that this would be a nice addition. Closing as a duplicate of #51.

Issue #186, opened 2020-01-09
Add support for Jupyter Notebooks


r0f1:
Feature request: A cool new feature would be to also crawl Jupyter Notebooks for import statements.

jonafato:
Agreed that this would be a nice addition. Closing as a duplicate of #51.

Issue #51, opened 2016-10-11
Jupyter Notebooks support


philippschw:
I just tried using pipreqs, but it seems useless for my workflow heavily relying on Jupyter notebooks. 

Will Jupyter notebooks be supported in the future?

bndr:
I'm sorry but, Jupyter will not be supported any time soon, I will leave this Issue open for future reference, if someday I find time to look into this.

nise:
Any updates here? I would love to use it.

DanielHabenicht:
Seems to be possible, someone wrote a wrapper: https://github.com/ivanlen/pipreqsnb

alan-barzilay:
implemented in #414

PR #210, opened 2020-07-02
Add support for jupyter notebook

pakio:
Related Issue
#51, #186, #199

Detail
Add support for jupyter notebook.
To parse notebook to python, used nbconvert library which officially provided by Project Jupyter.

note
As known issue, .ipynb which uses magic command without % cause Error when parsing script with awk.process().
For now I just left comment to let the user notice it, if you have good idea to fix it, I will be appreciated if you tell me.

----
Sorry for continue reopening PRs, I finally could find the reason why my coverage decreased and fixed it, also checked if the ci and coverall works fine in my repo.


pakio:
Hi @jamesmyatt cc @alan-barzilay , sorry for being stale for a long time. I just started to work on it again, but seeing an issue related to `nbconvert`. ([issue](https://github.com/jupyter/nbconvert/issues/1742), [failing test](https://github.com/pakio/pipreqs/runs/7374820152?check_suite_focus=true))
This could be resolved once I update the nbconvert version to latest or near-latest, but they are not supporting python 3.6 anymore.

Do you have any plan to stop supporting python3.6 in the near future?

jamesmyatt:
Im not a maintainer but Python 3.6 is EOL so dropping support makes sense.

Pereonally I'm in favour of small projects like this only supporting relatively few python versions for new releases. For example big projects often support 42 months of releases, like NEP29. But I would do 24 months.

PR #238, opened 2021-03-26
Mapping for github3

### LINKED ISSUE #226: pipreqs does not add github3.py to the requirements file
When pipreqs encounters `import github3` it does not know which package we plan to import.

After installing `pip install github3.py `to get github3 in the environment, then pipreqs gives both packages.

https://pypi.org/project/github3.py/

ryan-rozario/pipreqs-action#6

ryan-rozario:
@alan-barzilay  Added `github3:github3.py` to mapping file

Fixes #226


zkoppert:
Thank you @ryan-rozario !

PR #250, opened 2021-04-26
Get the correct version when passed multiple paths

solstag:
This emulates more closely the behavior of an actual import statement by:
1. traversing `sys.path` in reverse so we keep the first module found.
2. checking if there is actually a module at `path / i_item` to avoid false positives.

I've checked both behaviors in concrete cases and they both match the python
interpreter now, when previously they didn't.

An example where both 1. and 2. are necessary is when `sys.path` is
'/usr/lib:/usr/local/lib/pythonX.Y/site-packages/:/usr/lib/pythonX.Y/site-packages/'
and different versions of the package are installed in the last two paths.
Notice how `os.walk` will traverse '/lib/pythonX.Y/site-packages/' for the first path
('/usr/lib'), despite the fact that there's nothing to import from there, and that
traversing the paths in direct order would result in the last path overwriting the
version from the middle path, which comes first in import order.

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -209,7 +209,7 @@ def get_imports_info(
 def get_locally_installed_packages(encoding=None):
     packages = {}
     ignore = ["tests", "_tests", "egg", "EGG", "info"]
-    for path in sys.path:
+    for path in reversed(sys.path):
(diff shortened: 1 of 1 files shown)


solstag:
Ni! Looking at current code it seems this PR is no longer needed. Closing.

PR #253, opened 2021-05-07
Move docopt and yarg imports to local scope

befeleme:
I wonder, would you consider moving the imported libraries from global to local scope?

My motivation comes from pipenv which bundles pipreqs alongside with its dependencies (_yarg_ and _docopt_), but it really uses only two functions (`get_all_imports`, `get_pkg_names`) which don't require them.
When moved, pipenv could stop bundling _yarg_ and _docopt_.

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -42,10 +42,7 @@ import logging
 import codecs
 import ast
 import traceback
-from docopt import docopt
 import requests
-from yarg import json2package
-from yarg.exceptions import HTTPError
 
 from pipreqs import __version__
 
@@ -180,6 +177,10 @@ def output_requirements(imports):
 
 def get_imports_info(
         imports, pypi_server="https://pypi.python.org/pypi/", proxy=None):
+
+    from yarg import json2package
+    from yarg.exceptions import HTTPError
+
     result = []
 
     for item in imports:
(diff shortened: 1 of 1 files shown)


alan-barzilay:
@befeleme I'm not sure I understood your motivation, pipreqs is not a dependency of pipenv and I couldn't find anything that mentions pipreqs being bundled with pipenv. Could you provide a source regarding this?

befeleme:
Sure! See vendored dependencies of [pipenv](https://github.com/pypa/pipenv/tree/master/pipenv/vendor/pipreqs). It looks pipreqs is bundled alongside with its installable dependencies and used in [this](https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L254) function. The two dependencies mentioned in the PR are only pulled to make pipreqs import correctly. 

Regarding why - pipenv comes with so many bundled libraries, it's huge. I am currently testing pipenv's dependencies in Fedora, as we prefer to debundle and pull them directly. To be fair, removing two of them will not significantly reduce the size of pipenv, but it's feasible and reduces a drop of its complexity.

jonas-eschle:
Hi @befeleme , coming back to this issue, I think the only downside seems that anyone using the CLI would need to install now an extra, right?

befeleme:
Hi, feel free to just close this PR - pipreqs bundled dependency was removed from pipenv in 2022 (https://github.com/pypa/pipenv/pull/4999) and so I have no motivation to ask for changes in the project anymore.

jonas-eschle:
Great, thanks for letting us know and apologies for the late reply

PR #256, opened 2021-05-13
Fix 227

devashish-gaikwad:
I have added a new feature "--trusted-host" which disables the SSL verification of the requests sent by requests.get function call.
```
        --trusted-host        Ignore SSL warnings, recommended using with
                              enterprise proxy.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,6 @@ Session.vim
 *~
 
 /pipreqs/*.bak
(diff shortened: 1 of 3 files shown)


devashish-gaikwad:
Hello @alan-barzilay  !
I need your guidance here,

The travis builds which are failing are for pypy and

PR #257, opened 2021-05-13
Fix 227

devashish-gaikwad:
I have added a new feature "--trusted-host" which disables the SSL verification of the requests sent by requests.get function call.
```
        --trusted-host        Ignore SSL warnings, recommended using with
                              enterprise proxy.
```

Currently trying to fix pypy and flake8 build problems

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,6 @@ Session.vim
 *~
 
 /pipreqs/*.bak
+.vscode/settings.json
+.gitignore
+.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -55,4 +55,4 @@ Session.vim
 /pipreqs/*.bak
 .vscode/settings.json
 .gitignore
-.gitignore
+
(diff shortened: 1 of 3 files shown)


devashish-gaikwad:
Hi @alan-barzilay !

The build for Pypy seems to fail not because of the source code but because of some problem with the environment
Check out this build log: https://travis-ci.org/github/bndr/pipreqs/jobs/771058058
```
Could not build wheels for cryptography which use PEP 517 and cannot be installed directly
The command "pip install coverage coveralls tox" failed and exited with 1 during .
```

What do you think ?

alan-barzilay:
Hi @devashish-gaikwad, thank you for your PR! I will have to brush-up on SSL to understand if this is the best approach so it may take some time before I can get back to you. In the mean time, you should rebase this PR onto the `next` branch where we are focusing development. The pypy issue is also solved there :).

devashish-gaikwad:
Hi Alan! Thanks for the advice, I'll probably close this PR and open a new one with squashed commits. ( I don't really know how to rebase a PR from one place to another, I'll also look for a travis compatible testing way for proxy

Issue #252, opened 2021-05-01
pipreqs future


alan-barzilay:
Hi! I'm the new and (afaik) sole active maintainer of pipreqs. As you may have noticed, this package has been kinda abandoned in the last couple years. Well, I intend to change this and I would like to share my plan/vision with the community.

## Pipreqs current state

I've been closing a few issues and merging some PR's and I've noticed that most of them are quite old and/or duplicated. I've tried engaging with older issues but (understandably) I've been mostly ignored since people have probably moved on and don't remember or care about a 3 years old issue. Apart from these older issues there quite a few duplicated issues which can be  divided in feature requests (e.g. jupyter support) and mappings (e.g. package foobar is missing from my requirements).
The mapping problems are by far the majority of issues being reported, but attacking them feels like a wack-a-mole game where we just can't win and they just keep piling up.

## Where do we go from here?

The time I can dedicate to this project is limited since I have a thesis to write and other obligations in life, so instead of addressing the symptoms I would like to attack the source of our problems, i.e. our mapping system.

To be honest I'm not yet sure of the best way to address this, I think there are a few easy improvements that could be made but they are mostly band-aids, ideally we would completely remove the mapping file since this is a solution that does not scale very well. 
I'm open to suggestions on how to deal with this, but in the mean time I think I will make a PR template for adding package remapping and improve the documentation regarding this issue so users don't have to look at the source code to figure out why they have a missing package.

 These are my goals for the next release of pipreqs:

- [X] Add =~ and >= support

- [ ] Drop support for python 2

- [ ] Update pipreqs to python 3.9 (add missing standard lib packages, bump travis testing versions, etc)

- [ ] Add support for .ipynb and .pyw files 

- [ ] Add FAQ for common problems with package mapping

- [ ] Get rid of the mapping system (this may be pushed to a second release if this takes to long)

I will mostly dedicate my time to implement these changes, I won't be looking too much at the open issues and PR's made so I'm sorry if you end up being ignored. 

PS: there are a few open PR's that tackle some of these issues, but I have a few issues with them. I will contact the authors with my grievances and suggestions but some of them are quite old and I doubt they are still interested, nevertheless I will probably use them as a starting point.

rahul-kumi:
>ideally we would completely remove the mapping file since this is a solution that does not scale very well.

+1 to this suggestion.

alan-barzilay:
[...]

PR #272, opened 2021-09-04
Drop Travis as our CI provider

alan-barzilay:
Since travis changed its pricing policy it has become more limited in
what we can accomplish with it. It now uses a limited credit model for
open source projects which besides being cumbersome puts the project in
a precarious position where we may be unable to run tests if we dont
have any credits left.

For this reason we will be moving to github actions since they seem to
be the best alternative at the moment

I also dropped flake8 from our tox tests in favor of [reviewdog](https://github.com/reviewdog/action-flake8)
Review dog is an incredible project that makes linting and formatting
review a breeze by commenting inline what is wrong in a pull request.
This makes the review process easier for the maintainer and also
provides a clearer feedback to the contributor

This new test workflow utilizes a few interesting projects to make this action
easier to maintain such as the [coveralls github action](https://github.com/marketplace/actions/coveralls-github-action)
and the [tox-gh-actions project](https://github.com/ymyzk/tox-gh-actions)

This PR, as mentioned in #252, is necessary to get back our CI pipeline and to confidently make a new pipreqs release


alan-barzilay:
just a disclaimer that this is a work in progress, testing this CI is complicated because there is no way (afaik) of testing it locally.

I'm currently working on de-duplicating checks and investigating why some tests are being cancelled

alan-barzilay:
ok, so I cant find a way to keep using coveralls which sucks because we loose our coverage history. Uploading to coveralls without using the github action is really buggy, however the github action only supports reports in lcov format which i couldn't manage to create. See https://github.com/coverallsapp/github-action/issues/30

After 3 hours trying I gave up and tried codecov which just worked in 5 minutes. So unless someone finds a way to keep using coveralls, we will be using codecov from now on

alan-barzilay:
I tested this new CI with a dummy PR and everything seems to be working as expected! https://github.com/bndr/pipreqs/pull/273
This PR should be ready to be merged but I will wait a while in case anyone wants to give some feedback on this.

alan-barzilay:
also, our tests do not respect flake8, changing CI made this evident. I will be opening an issue about this

Issue #274, opened 2021-09-04
Tests do not respect flake8


alan-barzilay:
since we tested flake8 compliance with tox, the tests where never validated for flake8 violations

kuldeepkhatke:
Hi @alan-barzilay , 

I'm new to open source, I want to take up this issue. This will be my first issue. May I?

alan-barzilay:
Of course @kuldeepkhatke ! but just a heads up, I will be rebasing the "next" branch which may introduce some conflicts in your work. You said you were new to open source, do you know how to make a rebase and to resolve conflicts? If you dont you may want to wait a few days until I make the new pipreqs release and sort this all out

kuldeepkhatke:
@alan-barzilay , I'm new to open source but have good knowledge of git commands :)

alan-barzilay:
@kuldeepkhatke great! then just be warned that I will force push into the next branch. I need to make it inline with the master branch which has a revert commit to make it synchronized with the latest release. I will basically introduce the revert commit and then a revert of the revert.

(I know this is kinda dirty and that we should avoid force pushing but pipreqs needed some cleanup/book keeping and this was the cleanest way we could find. Once this is done we will have a clear and enforced development cicle and wont have any more force pushes, so please bear with me haha)

Iam-shyamrathore:
Hi @alan-barzilay is this still open

arnowaschk:
Is this still an issue in current code?

PR #276, opened 2021-09-07
Flake8 code quality #274

kuldeepkhatke:
Updated code formatting & long line break changes
Added warning ignore comments at unused imports, duplicate imports, multiple imports
Added line spaces after function definition

### DIFF
--- a/tests/_data/test.py
+++ b/tests/_data/test.py
@@ -1,32 +1,34 @@
 """unused import"""
-# pylint: disable=undefined-all-variable, import-error, no-absolute-import, too-few-public-methods, missing-docstring
+# pylint: disable=undefined-all-variable, import-error, no-absolute-import
+# pylint: disable=too-few-public-methods, missing-docstring
+from __future__ import print_function  # noqa:F401
 import xml.etree  # [unused-import]
-import xml.sax  # [unused-import]
-import os.path as test  # [unused-import]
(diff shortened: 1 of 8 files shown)


kuldeepkhatke:
Hi @alan-barzilay , 

Please review this PR & let me know if any changes needed.
Thanks

alan-barzilay:
hi @kuldeepkhatke, thank you for your contribution and welcome to pipreqs!

It seems like your changes broke one test, do you have any clue as to what may have caused this? From what I could gather you only added comments with directives for flake8, this shouldn't affect the unit tests. Did you change anything else that I missed?

Could you also add a comment in the top of the test file with a quick explanation about those directives and where someone could refer to learn more about them? those directives only make sense to someone already familiar with flake8, newcomers will likely get confused by them.
Btw, `F401` happens quite a lot, I think it would be better to just pass it as a code to be ignored directly to [flake8 in reviewdog](https://github.com/reviewdog/action-flake8)

Also, I think your 2 commits could be squashed into one since they all deal with general flake8 fixes.
Could you also change your merge into the "next" branch instead of master? I know they are synchronized right now, but I will make a new release pretty soon  and then this will no longer be the case

Sorry about the nitpicking and thanks again for your contribution!

jonas-eschle:
Can you maybe resolve the coflicts?

PR #277, opened 2021-09-13
Small change to make debugging file parse errors easier

abrahamw88:
Now when ast.parse() encounters a file that cannot be parsed due to syntax errors it will report that filename in it's traceback.


alan-barzilay:
hi @abrahamw88, thank you for your contribution and welcome to pipreqs!

> Now when ast.parse() encounters a file that cannot be parsed due to syntax errors it will report that filename in it's traceback.

Could you update your commit message with this? `update pipreqs.py` doesn´t speak much about what the commit does.

PR #279, opened 2021-09-16
Small change to make debugging file parse errors easier 2

abrahamw88:
Now when ast.parse() encounters a file that cannot be parsed due to syntax errors it will report that filename in it's traceback.


alan-barzilay:
hi @abrahamw88 thank you for the new pr! Are you familiar with [git rebase](https://git-scm.com/docs/git-rebase)? you could have done an interactive git rebase in your first PR in order to reword your commit. Another option would have been to use `git commit --amend`.

Could you use  git rebase to drop all commits besides the one with the actual changes? I understand that I may be sounding annoying by asking again for changes but this is important to keep the project organized and maintainable.
Also, when you write a commit message you should try to make the first line a simple and short (around 80 characters) description of what your commit does, then in the following lines you can enter into more details of what you did. 

One last tip: You can use a force push to alter this PR, this way you wont need to open a third PR. Normally force pushing is seen as something bad but since this is your personal branch that nobody else is using there is no problem in using a force push.

PR #287, opened 2021-11-19
Add --ignore-errors option

sify21:
This commit adds a new option --ignore-errors. Originally it's a
variable defined in get_all_imports and is always False. The purpose is
to ignore errors when parsing single .py file. Also the opening/reading
file part is moved to the try block, as those operations may also
fail(encoding problem).

Signed-off-by: 司芳源 <sify1221@gmail.com>


AliMostafaRadwan (approved):

pipreqs/pipreqs.py
```diff
@@ -111,9 +111,9 @@ def get_all_imports(
...
         for file_name in files:
             file_name = os.path.join(root, file_name)
-            with open(file_name, "r", encoding=encoding) as f:
-                contents = f.read()
             try:
+                with open(file_name, "r", encoding=encoding) as f:
+                    contents = f.read()
                 tree = ast.parse(contents)
                 for node in ast.walk(tree):
                     if isinstance(node, ast.Import):
```
jonas-eschle, line 119: I think there is still a bug here: `traceback.print_exc(exc)` (on line 126 just below) doesn't take the exception as argument, but the limit. So it should be `traceback.print_exc()` without an argument

jonas-eschle (approved):
Looks also fine to me, just wanted to make a PR with the same content. When will this get merged?

pipreqs/pipreqs.py
sify21: image
There are api changes in version 3.5 and 3.10

jonas-eschle:
@AliMostafaRadwan can we get this merged? Seems ready and a useful change

jonas-eschle:
@alan-barzilay this seems ready, can we get this merged?

jonas-eschle:
@sify21 can you resolve the conflicts? Or are there any blockers?

sify21:
@jonas-eschle resolved

jonas-eschle:
@alan-barzilay can me maybe merge this?

jonas-eschle:
There is another, maybe more extensive PR here: https://github.com/bndr/pipreqs/pull/473

Can you maybe comment there and tell what could be refined?

PR #294, opened 2022-01-13
Please allow using this project a library

oz123:
This project is very useful also as a library. For example, pipenv uses it here:

https://github.com/pypa/pipenv/blob/4fa67914f7562db947e1b30875cf96a709e8f4f5/pipenv/core.py#L219

Because of the top level import of docopt, it must also pull docopt as a dependecy. It would be really nice if we
could remove this dependency.

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -44,7 +44,6 @@ import re
 import logging
 import ast
 import traceback
-from docopt import docopt
 import requests
 from yarg import json2package
 from yarg.exceptions import HTTPError
@@ -602,6 +601,11 @@ def init(args):
 
 
 def main():  # pragma: no cover
+    try:
+        from docopt import docopt
+    except (ImportError, ModuleNotFoundError):
+        sys.exit("Please install docopt to use the command line")
+
     args = docopt(__doc__, version=__version__)
     log_level = logging.DEBUG if args["--debug"] else logging.INFO
     logging.basicConfig(level=log_level, format="%(levelname)s: %(message)s")
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,6 @@ classifiers = [
 requires-python = ">=3.9, <3.14"
 dependencies = [
     "yarg>=0.1.9",
-    "docopt>=0.6.2",
     "nbconvert>=7.11.0",
     "ipython>=8.12.3",
 ]
@@ -39,6 +38,10 @@ dev = [
     "coverage>=7.3.2",
     "sphinx>=7.2.6;python_version>='3.9'",
 ]
+cli = [
+    "docopt>=0.6.2",
+]
+
 [tool.poetry.group.dev.dependencies]  # for legacy usage
 flake8 = "^6.1.0"
 tox = "^4.11.3"


oz123:
Just noticed that there is another PR that suggested this:

https://github.com/bndr/pipreqs/pull/253.

However, it regards two dependencies, and does not modify `setup.py`. 
Let's start with this one, first remove docopt, and later yarg.

jonas-eschle:
Hi, thanks a lot for the PR, this seems reasonable to me. Can you resolve the conflicts and change things in the updated structure? I.e. the pyproject.toml?

jonas-eschle:
I think the main issue I see with this is that it will be a breaking change, right? I'm similarly unease to have ipython as a dependency tbh

StantonMatt (commented):
I took a local pass on this because the library-use motivation is still valid, especially for downstreams that import `pipreqs.pipreqs` without needing the CLI parser.

The import-time part works as intended for me. In a clean default editable install, `import pipreqs.pipreqs` succeeds and `docopt` is not installed:

```console
UV_CACHE_DIR=.codex-tmp/uv-cache uv pip install --python .codex-tmp/venvs/pipreqs-294-default/bin/python -e repos/pipreqs
.codex-tmp/venvs/pipreqs-294-default/bin/python - <<'PY'
import importlib.util
import pipreqs.pipreqs
print('library import ok')
print('docopt present:', importlib.util.find_spec('docopt') is not None)
PY
```

That prints `library import ok` and `docopt present: False`.

I think there are two release/package details to handle before this is merge-ready:

1. `poetry.lock` needs to be regenerated. On the PR branch, `poetry check --lock` reports:

```console
Error: pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock` to fix the lock file.
```

The lock file still has a `docopt` package entry and the old content hash.

2. This makes the default console-script install a breaking change for CLI users. With the default install, `pipreqs --version` now exits 1 with `Please install docopt to use the command line`; with `pipreqs[cli]`, it works. That may be fine for the next major release, but I think the README/install docs or release notes should make the new `pipreqs[cli]` path explicit so users do not upgrade into a broken CLI unexpectedly.

Other checks I ran:

```console
UV_CACHE_DIR=.codex-tmp/uv-cache uv pip install --python .codex-tmp/venvs/pipreqs-294-cli/bin/python -e 'repos/pipreqs[cli,dev]'
.codex-tmp/venvs/pipreqs-294-cli/bin/pipreqs --version
.codex-tmp/venvs/pipreqs-294-cli/bin/python -m unittest tests.test_pipreqs.TestPipreqs.test_get_pkg_names tests.test_pipreqs.TestPipreqs.test_get_use_local_only tests.test_pipreqs.TestPipreqs.test_ignore_notebooks
.codex-tmp/venvs/pipreqs-294-cli/bin/python -m py_compile pipreqs/pipreqs.py tests/test_pipreqs.py
.codex-tmp/venvs/pipreqs-294-cli/bin/python -m flake8 pipreqs tests
```

Those passed.

Full `python -m unittest discover` fails in this environment at `TestPipreqs.test_init` because the generated requirements include both `ipython==8.18.1` and `ipython==8.12.3`, which breaks the existing sorted-output assertion. I checked the same focused failure on current `origin/master`, and it reproduces there too, so I would treat that as a baseline/environment issue rather than a regression from this PR.

Issue #265, opened 2021-06-25
Pipreqs reports wrong version when package name != module name


aaron-santos-agrian:
Steps to reproduce:
```bash
python3 -m venv testenv
source testenv/bin/activate
# install older version of scikit-learn
pip install pipreqs scikit-learn==0.20.3
mkdir test
echo "import sklearn" > test/test.py
pipreqs test
cat test/requirements.txt
deactivate
```
Actual Output
`scikit_learn==0.24.2`

Expected Output
`scikit_learn==0.20.3`

This occurs because `get_import_local` is called with package names in the `imports` parameter (by way of `candidates = get_pkg_names(candidates)`). It tests if each import is in `local`. However, `local` is a dict where the keys are module names. This results in not finding the locally installed package and instead looking up the latest version on pypi.org.

PR #299, opened 2022-02-14
Update pipreqs.py

janschmid:
Hello, i came across the same error as described here and made a quick fix: 
https://github.com/bndr/pipreqs/issues/265

Seems to work in my case, tested it with python 3.6.16, and i don't see why it should conflict with other versions


jonas-eschle:
Hey @janschmid thanks for this! Could you add a quick test maybe with a package?

PR #310, opened 2022-05-23
Update README.rst

somyagr:
Corrected the command line


jonas-eschle:
Mh, that seems more like changing an absolute path to a relative path, right? Not sure this should be changed.

PR #316, opened 2022-07-21
added apparent_encoding to the decoding of pypi's json response

mglac:
The json response from pypi is not guaranteed to be all of the expected default encoding. When characters in a pypi package information are not the default encoding, this line here will error out. By using `response.apparent_encoding`, the encoding type is taken from the request's response to allow for a proper decode.

There are a lot of repositories that cause issues like this, but here is an example:
https://pypi.org/project/imath/


jonas-eschle:
Hi, thanks for the PR! Could you add a test to this?

PR #325, opened 2022-09-06
Doc: no-pin instead of non-pin for scheme

klieret:
If there is no specific reason to call it `non-pin` in the explanation text and `no-pin` for the option, I would change the explanation text:

```
    --mode <scheme>       Enables dynamic versioning with <compat>, <gt> or <non-pin> schemes
                          <compat> | e.g. Flask~=1.1.2
                          <gt>     | e.g. Flask>=1.1.2
                          <no-pin> | e.g. Flask
```


jonas-eschle:
Thanks @klieret, old but good ;) LGTM!

Issue #326, opened 2022-09-29
UnicodeDecodeError: 'gbk' codec can't decode byte 0xa1 in position 3608: illegal multibyte sequence


code4happylife:
After I run   
```

pipreqs . 
```
I got errors like below:

File "C:\Users\myname\AppData\Local\Programs\Python\Python310\lib\site-packages\pipreqs\pipreqs.py", line 115, in get_all_imports
    contents = f.read()

UnicodeDecodeError: 'gbk' codec can't decode byte 0xa1 in position 3608: illegal multibyte sequence

ConyKitty:
Try to add the parameter `--encoding` with UTF-8:

`pipreqs . --encoding UTF-8`

alan-barzilay:
fixed in #414

PR #327, opened 2022-09-29
update encoding setting as UTF-8

code4happylife:
update encoding setting as UTF-8 as well as solve the issue:https://github.com/bndr/pipreqs/issues/326


alan-barzilay:
implemented in #414

PR #332, opened 2022-10-20
Bump minor version + correct encoding errors people face with cp1252

Moooog:
Bump minor version to 0.4.12 + correct encoding errors people face with cp1252 (cp850 works but not recommended)


jonas-eschle:
Should we just ignore errors by default?

PR #339, opened 2022-12-02
Require Python 3.7 or higher

EwoutH:
Require Python 3.7 by setting `python_requires` in setup.py. Also updates the CI, tests and docs.


alan-barzilay:
for some reason i can approve it here haha

PR #348, opened 2023-01-13
Tox test

c-w-m:
fix_SSLError final
- implementation of new arg (--verify <CA_BUNDLE> to allow requests.get setting of verify arg
- testing updates
  - dotnet used with testing to allow `.env.test` file for setting CA_BUNDLE environment variable
  - new `settings.py` file used in testing to read in either environment variable or .env.test variable
  - tox file revised to now include py3.10 and a development build of py3.11
  -  /.github/workflow/tests.yml were not updated (they still have `python-version: [3.7, 3.8, 3.9, pypy-3.7]`)
- doc updates
  - README.rst has new `--verify` option 
  - CONTRIBUTING.rst note that 3.7 to 3.10 should pass (3.11 needs work)


jonas-eschle:
Thanks also for this PR! What do you think, should we still add this?

PR #351, opened 2023-01-20
CA_BUNDLE environment variable is working now

c-w-m:
Previous testing did not uncover an issue with the environment variable string format.  This version is not passing the tests and use of the CA_BUNDLE environment variable.


jonas-eschle:
Hi there, thanks for the PR! Is this still needed?

PR #369, opened 2023-04-18
Update README.rst

Muhammad-Aadil:
Wrap the PIP Installation in the code block so that it will be easy to copy and run the command in the shell. 

Now a nice "copy to clipboard" icon appears.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@
 Installation
 ------------
 
-::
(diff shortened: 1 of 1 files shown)


alan-barzilay:
thanks for the contribution!
cherry-picked into PR https://github.com/bndr/pipreqs/pull/402

PR #377, opened 2023-06-16
Add --extra option

Temerold:
By using the --extra option, users can add custom packages to the `requirements.txt` file, like so:
`$ pipreqs --extra pipreqs,example`
This command generates the correct file, alphabetized and all, including pipreqs and example (in this case).

This adds pipreqs itself and example to the `requirements.txt` file, as my original intention was to add pipreqs to the `requirements.txt` file. (I know there are severely easier ways to do so, but I implemented it anyway.)

I added the `test_extra_module()` test, which it passed alongside all other tests; should be good to go!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,6 +46,7 @@ Usage
                               $ export HTTPS_PROXY="https://10.10.1.10:1080"
         --debug               Print debug information
         --ignore <dirs>...    Ignore extra directories, each separated by a comma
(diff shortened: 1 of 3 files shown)


Borda (approved):
this would be great addition

Temerold:
> this would be great addition

Thank you!

This repo seems however to be very inactive, and I'm thinking about forking and making my own pip package at this point....

PR #390, opened 2023-09-20
Add unit test for "get_locally_installed_packages" function

willianrocha:
validating ignore module and packages

### DIFF
--- a/tests/test_pipreqs.py
+++ b/tests/test_pipreqs.py
@@ -11,6 +11,8 @@ Tests for `pipreqs` module.
 import io
 import sys
 import unittest
+from unittest.mock import patch
+import sys
 import os
 import requests
 
@@ -48,7 +50,14 @@ class TestPipreqs(unittest.TestCase):
             os.path.dirname(__file__), "_data_duplicated_deps"
         )
         self.requirements_path = os.path.join(self.project, "requirements.txt")
-        self.alt_requirement_path = os.path.join(self.project, "requirements2.txt")
+        self.alt_requirement_path = os.path.join(
+            self.project,
+            "requirements2.txt"
+            )
+        self.local_packages_path = os.path.join(
+            os.path.dirname(__file__),
+            "_local_packages"
+            )
 
     def test_get_all_imports(self):
         imports = pipreqs.get_all_imports(self.project)
@@ -108,6 +117,15 @@ class TestPipreqs(unittest.TestCase):
         imports_with_info = pipreqs.get_import_local(self.modules)
         for item in imports_with_info:
             self.assertTrue(item["name"].lower() in self.local)
+
+    def test_get_locally_installed_packages(self):
+        with patch.object(sys, 'path', new=[self.local_packages_path]):
+            local_installed_packages = pipreqs.get_locally_installed_packages()
+            expected_output = [
+                {'exports': ['valid_package'], 'name': 'valid_package', 'version': '1.0.0'},
+                {'exports': [], 'name': 'EGG', 'version': 'INFO'}
+            ]
+            self.assertEqual(local_installed_packages, expected_output)
 
     def test_init(self):
         """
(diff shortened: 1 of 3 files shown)


tests/test_pipreqs.py
```diff
@@ -203,7 +219,7 @@ def test_ignored_directory(self):
         Test --ignore parameter
         """
         pipreqs.init(
-            {'<path>': self.project_with_ignore_directory, '--savepath': None, 
+            {'<path>': self.project_with_ignore_directory, '--savepath': None,
```
alan-barzilay, line 222: this changes (probably made by your IDE) only muddles the commit, could you remove them?

tests/test_pipreqs.py
```diff
@@ -109,6 +115,16 @@ def test_get_use_local_only(self):
         for item in imports_with_info:
             self.assertTrue(item['name'].lower() in self.local)
 
+    def test_get_locally_installed_packages(self):
+        with patch.object(sys, 'path', new=[self.local_packages_path]):
+            local_installed_packages = pipreqs.get_locally_installed_packages()
+            expected_output = [
+                {'exports': [], 'name': 'invalid_package', 'version': None},
+                {'exports': ['valid_package'], 'name': 'valid_package', 'version': '1.0.0'},
+                {'exports': [], 'name': 'EGG', 'version': 'INFO'}
```
alan-barzilay, line 124: sorry, I dont understand this test. Could you explain it to me? why this becomes the output and why is the module invalid? Is it because it haz an additional folder?

alan-barzilay (changes requested):

alan-barzilay:
Also, you should push to next, not master

alan-barzilay:
@willianrocha could you take a look at PR #384 also? It seems related to this test, I would like your opinion about the issue, are non top_level pip packages something compliant with  how pip works and the documentation? if so it seems one of the test cases implemented in this test could be a problem

tests/test_pipreqs.py
mateuslatrova: First of all, we based our tests on our understanding of the code. We don't know exactly how pip builds the structure of a package.

1. The first test case was not necessary, so we decided to remove it.
2. The second one was called as a "valid_package" because its "top_level.txt" file has one module named "valid_package" that is not in our code's blacklist (which is a variable called ignore with value `["tests", "_tests", "egg", "EGG", "info"]`) and also has a module named "tests" that is in that blacklist.
3. The last test case makes sure that "top_level.txt" files that are inside a folder that starts with "EGG" are not considered.

We tried to understand the code's behavior by reading "setuptools" documentation ([link](https://setuptools.pypa.io/en/latest/deprecated/python_eggs.html)), but we did not understand much. It requires a lot of reading and time for understanding.

If you think it's okay for us to use our time to study this and then validate if it's right, that is okay.

mateuslatrova:
> @willianrocha could you take a look at PR #384 also? It seems related to this test, I would like your opinion about the issue, are non top_level pip packages something compliant with how pip works and the documentation? if so it seems one of the test cases implemented in this test could be a problem

We looked into some packages that are installed in our environments. 

Packages that are in the "EGG" format have the "top_level.txt" file, but others don't. For example, pandas is a package in "dist.info" format and it does not have the "top_level.txt" file, but it has the "METADATA" file mentioned in the PR you just said. 

I don't know which one is compliant or not, but it seems to me that maybe our code is not getting all information it is looking for... In the case of pandas, for example, it would not get the package from the "top_level.txt" and would continue running...

Also, in the EGG format, the "PKG-INFO" file seems a lot like the "METADATA" file from "DIST-INFO" format.

Obs.: it looks like EGG packages are deprecated...

Maybe we could handle both types of distribution...

tests/test_pipreqs.py
```diff
@@ -11,6 +11,8 @@
 import io
 import sys
 import unittest
+from unittest.mock import patch
+import sys
```
alan-barzilay, line 15: sys is already imported in line 12, 3 lines up haha

alan-barzilay (changes requested):

PR #382, opened 2023-08-02
Handle everything in lowercase [issue #381]

wurmmi:
fixes issue https://github.com/bndr/pipreqs/issues/381

Issue #381, opened 2023-08-02
pipreqs should not be case sensitive


wurmmi:
We should handle all package names in lowercase, since **pip is case insensitive**.
- see PEP-426 (https://peps.python.org/pep-0426/#name)
- see PEP-508 (https://peps.python.org/pep-0508/#names)

This will also reduce the amount of mappings. Many of them only match the names to upper-/lowercase.

PR #395, opened 2023-09-21
create test for output_requirements

fredgrub:


tests/test_pipreqs.py
```diff
@@ -322,6 +324,30 @@ def test_clean_with_imports_to_clean(self):
...
+            '--use-local': None, '--force': None, '--proxy': None,
+            '--pypi-server': None, '--diff': None, '--clean': None,
+            '--mode': None}
+            )
+        pipreqs.init({'<path>': self.project, '--savepath': None, '--print': False,
+                      '--use-local': None, '--force': True, '--proxy':None, '--pypi-server':None,
+                      '--diff': None, '--clean': None, '--mode': None})
+        
+        with open(self.requirements_path, "r") as f:
+            file_content = f.read().lower()
```
alan-barzilay, line 347: remove lower, should be unnecessary

alan-barzilay (changes requested):
Dont forget to rebase your PR

tests/test_pipreqs.py
fredgrub: Solved on [321cb8b](https://github.com/bndr/pipreqs/pull/395/commits/321cb8b45ebc340d3d5395c4fe236169319bbe64).

alan-barzilay: thanks @fredgrub ! Last thing about this lower issue, could you reword the commit?

Give some more context about the issue and its solution, if someone reads "remove lower" as a commit message they wont have any idea as to what the commit does or why it was necessary, they will need to go to the commit changes and try to decipher what it means and why it was necessary.

There is no need to write a whole essay but some additional context and it information would be good

tests/test_pipreqs.py
```diff
@@ -427,6 +429,30 @@ def test_clean_with_imports_to_clean(self):
...
+        Test --print parameter
+        It should print to stdout the same content as requeriments.txt 
+        """
+
+        capturedOutput = io.StringIO()          
+        sys.stdout = capturedOutput
+
+        pipreqs.init(
+            {'<path>': self.project, '--savepath': None, '--print': True,
+            '--use-local': None, '--force': None, '--proxy': None,
```
alan-barzilay, line 443: @fredgrub, the flake8 issues dont seem to have dissapeared even after rebasing. Could you take a look at them? And its weird that it passed the flake8 test while giving a bunch of warnings and erors...

tests/test_pipreqs.py
alan-barzilay: sorry for being a pain in the ass haha but #381 #382 raise a fair point... Do you have any thoughts about it? maybe it would be better to just drop this "lower" commit from this PR and create a new task for issue #381

fredgrub: I think that the issue of case sensitivity should be moved to a separate pull request. I just dropped the "remove lower" commit.

tests/test_pipreqs.py
fredgrub: I forgot to format the modifications done in this PR. The flake8 was only applied to the other lines.

fredgrub:
I just updated the PR description with more details.

alan-barzilay (approved):

PR #404, opened 2023-10-09
Add support for jupyter notebook

Fernando-crz:
- Commit rebased from https://github.com/bndr/pipreqs/pull/210 . 
- Credits to pakio: https://github.com/pakio

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,6 +40,7 @@ jobs:
               run: |
                   python -m pip install --upgrade pip
                   pip install coverage docopt yarg requests nbconvert
+                  pip install .[jupyter]
 
             - name: Calculate coverage
               run: coverage run --source=pipreqs -m unittest discover
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,7 +39,7 @@ jobs:
             - name: Install dependencies
               run: |
                   python -m pip install --upgrade pip
-                  pip install coverage docopt yarg requests
+                  pip install coverage docopt yarg requests nbconvert
 
             - name: Calculate coverage
               run: coverage run --source=pipreqs -m unittest discover
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -170,10 +170,6 @@ def get_all_imports(path, encoding=None, extra_ignore_dirs=None, follow_links=Tr
     return list(packages - data)
 
 
-def filter_line(line):
-    return len(line) > 0 and line[0] != "#"
-
-
 def filter_ext(file_name, acceptable):
     return os.path.splitext(file_name)[1] in acceptable
 
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -47,17 +47,18 @@ from docopt import docopt
 import requests
 from yarg import json2package
 from yarg.exceptions import HTTPError
+from nbconvert import PythonExporter
 
 from pipreqs import __version__
 
 REGEXP = [
-    re.compile(r'^import (.+)$'),
-    re.compile(r'^from ((?!\.+).*?) import (?:.*)$')
+    re.compile(r"^import (.+)$"),
+    re.compile(r"^from ((?!\.+).*?) import (?:.*)$"),
 ]
 
 
 @contextmanager
-def _open(filename=None, mode='r'):
+def _open(filename=None, mode="r"):
     """Open a file or ``sys.stdout`` depending on the provided filename.
 
     Args:
@@ -70,13 +71,13 @@ def _open(filename=None, mode='r'):
         A file handle.
 
     """
-    if not filename or filename == '-':
-        if not mode or 'r' in mode:
+    if not filename or filename == "-":
+        if not mode or "r" in mode:
             file = sys.stdin
-        elif 'w' in mode:
+        elif "w" in mode:
             file = sys.stdout
         else:
-            raise ValueError('Invalid mode for file: {}'.format(mode))
+            raise ValueError("Invalid mode for file: {}".format(mode))
     else:
         file = open(filename, mode)
 
@@ -87,13 +88,21 @@ def _open(filename=None, mode='r'):
             file.close()
 
 
-def get_all_imports(
-        path, encoding=None, extra_ignore_dirs=None, follow_links=True):
+def get_all_imports(path, encoding=None, extra_ignore_dirs=None, follow_links=True):
     imports = set()
     raw_imports = set()
     candidates = []
     ignore_errors = False
-    ignore_dirs = [".hg", ".svn", ".git", ".tox", "__pycache__", "env", "venv"]
+    ignore_dirs = [
+        ".hg",
+        ".svn",
+        ".git",
+        ".tox",
+        "__pycache__",
+        "env",
+        "venv",
+        ".ipynb_checkpoints",
+    ]
 
     if extra_ignore_dirs:
         ignore_dirs_parsed = []
@@ -106,13 +115,23 @@ def get_all_imports(
         dirs[:] = [d for d in dirs if d not in ignore_dirs]
 
         candidates.append(os.path.basename(root))
-        files = [fn for fn in files if os.path.splitext(fn)[1] == ".py"]
+        files = [fn for fn in files if filter_ext(fn, [".py", ".ipynb"])]
 
-        candidates += [os.path.splitext(fn)[0] for fn in files]
+        candidates = list(
+            map(
+                lambda fn: os.path.splitext(fn)[0],
+                filter(lambda fn: filter_ext(fn, [".py"]), files),
+            )
+        )
+
         for file_name in files:
             file_name = os.path.join(root, file_name)
-            with open(file_name, "r", encoding=encoding) as f:
-                contents = f.read()
+            contents = ""
+            if filter_ext(file_name, [".py"]):
+                with open(file_name, "r", encoding=encoding) as f:
+                    contents = f.read()
+            elif filter_ext(file_name, [".ipynb"]):
+                contents = ipynb_2_py(file_name, encoding=encoding)
             try:
                 tree = ast.parse(contents)
                 for node in ast.walk(tree):
@@ -128,6 +147,8 @@ def get_all_imports(
                     continue
                 else:
                     logging.error("Failed on file: %s" % file_name)
+                    if filter_ext(file_name, [".ipynb"]):
+                        logging.error("Magic command without % might be failed")
                     raise exc
 
     # Clean up imports
@@ -137,11 +158,11 @@ def get_all_imports(
         # Cleanup: We only want to first part of the import.
         # Ex: from django.conf --> django.conf. But we only want django
         # as an import.
-        cleaned_name, _, _ = name.partition('.')
+        cleaned_name, _, _ = name.partition(".")
         imports.add(cleaned_name)
 
     packages = imports - (set(candidates) & imports)
-    logging.debug('Found packages: {0}'.format(packages))
+    logging.debug("Found packages: {0}".format(packages))
 
     with open(join("stdlib"), "r") as f:
         data = {x.strip() for x in f}
@@ -149,58 +170,81 @@ def get_all_imports(
     return list(packages - data)
 
 
+def filter_line(line):
+    return len(line) > 0 and line[0] != "#"
+
+
+def filter_ext(file_name, acceptable):
+    return os.path.splitext(file_name)[1] in acceptable
+
+
+def ipynb_2_py(file_name, encoding=None):
+    """
+
+    Args:
+        file_name (str): notebook file path to parse as python script
+        encoding  (str): encoding of file
+
+    Returns:
+        str: parsed string
+
+    """
+
+    exporter = PythonExporter()
+    (body, _) = exporter.from_filename(file_name)
+
+    return body.encode(encoding if encoding is not None else "utf-8")
+
+
 def generate_requirements_file(path, imports, symbol):
     with _open(path, "w") as out_file:
-        logging.debug('Writing {num} requirements: {imports} to {file}'.format(
-            num=len(imports),
-            file=path,
-            imports=", ".join([x['name'] for x in imports])
-        ))
-        fmt = '{name}' + symbol + '{version}'
-        out_file.write('\n'.join(
-            fmt.format(**item) if item['version'] else '{name}'.format(**item)
-            for item in imports) + '\n')
+        logging.debug(
+            "Writing {num} requirements: {imports} to {file}".format(
+                num=len(imports),
+                file=path,
+                imports=", ".join([x["name"] for x in imports]),
+            )
+        )
+        fmt = "{name}" + symbol + "{version}"
+        out_file.write(
+            "\n".join(fmt.format(**item) if item["version"] else "{name}".format(**item) for item in imports) + "\n"
+        )
 
 
 def output_requirements(imports, symbol):
-    generate_requirements_file('-', imports, symbol)
+    generate_requirements_file("-", imports, symbol)
 
 
-def get_imports_info(
-        imports, pypi_server="https://pypi.python.org/pypi/", proxy=None):
+def get_imports_info(imports, pypi_server="https://pypi.python.org/pypi/", proxy=None):
     result = []
 
     for item in imports:
         try:
             logging.warning(
-                'Import named "%s" not found locally. '
-                'Trying to resolve it at the PyPI server.',
-                item
+                'Import named "%s" not found locally. ' "Trying to resolve it at the PyPI server.",
+                item,
             )
-            response = requests.get(
-                "{0}{1}/json".format(pypi_server, item), proxies=proxy)
+            response = requests.get("{0}{1}/json".format(pypi_server, item), proxies=proxy)
             if response.status_code == 200:
-                if hasattr(response.content, 'decode'):
+                if hasattr(response.content, "decode"):
                     data = json2package(response.content.decode())
                 else:
                     data = json2package(response.content)
             elif response.status_code >= 300:
-                raise HTTPError(status_code=response.status_code,
-                                reason=response.reason)
+                raise HTTPError(status_code=response.status_code, reason=response.reason)
         except HTTPError:
-            logging.warning(
-                'Package "%s" does not exist or network problems', item)
+            logging.warning('Package "%s" does not exist or network problems', item)
             continue
         logging.warning(
             'Import named "%s" was resolved to "%s:%s" package (%s).\n'
-            'Please, verify manually the final list of requirements.txt '
-            'to avoid possible dependency confusions.',
+            "Please, verify manually the final list of requirements.txt "
+            "to avoid possible dependency confusions.",
             item,
             data.name,
             data.latest_release_id,
-            data.pypi_url
+            data.pypi_url,
         )
-        result.append({'name': item, 'version': data.latest_release_id})
+        result.append({"name": item, "version": data.latest_release_id})
     return result
 
 
@@ -225,25 +269,23 @@ def get_locally_installed_packages(encoding=None):
                         filtered_top_level_modules = list()
 
                         for module in top_level_modules:
-                            if (
-                                (module not in ignore) and
-                                (package[0] not in ignore)
-                            ):
+                            if (module not in ignore) and (package[0] not in ignore):
                                 # append exported top level modules to the list
                                 filtered_top_level_modules.append(module)
 
                         version = None
                         if len(package) > 1:
-                            version = package[1].replace(
-                                ".dist", "").replace(".egg", "")
+                            version = package[1].replace(".dist", "").replace(".egg", "")
 
                         # append package: top_level_modules pairs
                         # instead of top_level_module: package pairs
-                        packages.append({
-                            'name': package[0],
-                            'version': version,
-                            'exports': filtered_top_level_modules
-                        })
+                        packages.append(
+                            {
+                                "name": package[0],
+                                "version": version,
+                                "exports": filtered_top_level_modules,
+                            }
+                        )
     return packages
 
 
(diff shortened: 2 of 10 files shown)


alan-barzilay:
tests failed due to nbconvert version, have you run the tests locally?
Also, have you addressed the issue raised by @jamesmyatt about optional dependency?

jamesmyatt:
I think @alan-barzilay means this one: https://github.com/bndr/pipreqs/pull/210#discussion_r734564597

Fernando-crz:
Just fixed bugs related to the tests. The tests were failing because they were not considering that ipython was now part of the locally installed packages. Just fixed this issue now and tests are running just fine, but for some odd reason the codecov coverage report failed. @alan-barzilay

Fernando-crz:
Just added support for jupyter notebook as an optional depency, as adressed by https://github.com/bndr/pipreqs/pull/210#discussion_r734564597.
@alan-barzilay

pipreqs/pipreqs.py
```diff
@@ -473,20 +513,21 @@ def init(args):
...
-                      x.lower() not in [y for x in local for y in x['exports']]
-                      and
-                      # check if candidate is package names
-                      x.lower() not in [x['name'] for x in local]]
-
-        imports = local + get_imports_info(difference,
-                                           proxy=proxy,
-                                           pypi_server=pypi_server)
+        difference = [
+            x
```
alan-barzilay, line 517: formatters are weird

setup.py
```diff
@@ -15,7 +15,11 @@
     history = history_file.read().replace('.. :changelog:', '')
 
 requirements = [
-    'docopt', 'yarg', 'nbconvert', 'ipython'
+    'docopt', 'yarg'
+]
+
+jupyter_requirements = [
```
alan-barzilay, line 21: please eli5 how this works or link a reference haha

setup.py
```diff
@@ -34,6 +38,7 @@
     include_package_data=True,
     package_data={'': ['stdlib', 'mapping']},
     install_requires=requirements,
+    extras_require={"jupyter":jupyter_requirements},
```
alan-barzilay, line 41: woof woof

alan-barzilay (changes requested):
guys, you gotta pay closer attention when you are using someone else's commit haha  even more so when they are that old and things have diverged quite a bii. Its ok to pass them through a formatter or to create a fixup  to alter part of their code (i believe that would give both of you authorship of the commit, and thats ok)

pipreqs/pipreqs.py
```diff
@@ -47,17 +47,18 @@
 import requests
 from yarg import json2package
 from yarg.exceptions import HTTPError
+from nbconvert import PythonExporter
 
 from pipreqs import __version__
 
 REGEXP = [
-    re.compile(r'^import (.+)$'),
```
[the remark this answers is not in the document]
Fernando-crz: Since the repository is using double quotes as standard, i will just leave it as it is

pipreqs/pipreqs.py
```diff
@@ -137,70 +158,93 @@ def get_all_imports(
...
+
+    Returns:
+        str: parsed string
+
+    """
+
+    exporter = PythonExporter()
+    (body, _) = exporter.from_filename(file_name)
+
+    return body.encode(encoding if encoding is not None else "utf-8")
```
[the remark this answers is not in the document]
Fernando-crz: done

pipreqs/pipreqs.py
```diff
@@ -137,70 +158,93 @@ def get_all_imports(
...
-    logging.debug('Found packages: {0}'.format(packages))
+    logging.debug("Found packages: {0}".format(packages))
 
     with open(join("stdlib"), "r") as f:
         data = {x.strip() for x in f}
 
     return list(packages - data)
 
 
+def filter_line(line):
```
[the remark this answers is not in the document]
Fernando-crz: Done

pipreqs/pipreqs.py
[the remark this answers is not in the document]
Fernando-crz: Setting the default value to utf-8 makes the tests fail. It happens because, when calling the `init` function without an encoding option, it defaults to _None_, passing _None_ as an argument to to the `get_all_imports` function, which passes _None_ as the encoding argument to the `ipynb_2_py` function. Leaving the function working in this way, makes it capable of dealing with _None_ values as input without crashing the program.

pipreqs/pipreqs.py
```diff
@@ -256,14 +298,14 @@ def get_import_local(imports, encoding=None):
...
-            if item in package['exports'] or item == package['name']:
+            if item in package["exports"] or item == package["name"]:
                 result.append(package)
 
     # removing duplicates of package/version
     # had to use second method instead of the previous one,
     # because we have a list in the 'exports' field
     # https://stackoverflow.com/questions/9427163/remove-duplicate-dict-in-list-in-python
-    result_unique = [i for n, i in enumerate(result) if i not in result[n+1:]]
+    result_unique = [i for n, i in enumerate(result) if i not in result[n + 1 :]]
```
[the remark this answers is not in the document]
Fernando-crz: Done

pipreqs/pipreqs.py
Fernando-crz: What should we do in this case? Even if I solved this problem manually fixing this formater weirdness, next time someone runs the formatter it will just go back to this weird thing...

setup.py
Fernando-crz: Here's a reference on how it works: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

setup.py
Fernando-crz: Done @alan-barzilay

pipreqs/pipreqs.py
```diff
@@ -128,6 +147,8 @@ def get_all_imports(
                     continue
                 else:
                     logging.error("Failed on file: %s" % file_name)
+                    if filter_ext(file_name, [".ipynb"]):
+                        logging.error("Magic command without % might be failed")
```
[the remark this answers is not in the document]
Fernando-crz: Basically, jupyter notebooks have this feature called _Magic commands_, which give new possible commands to our program using IPython support. Magic commands can be called using the "%" sign, but there's a way to call those commands without using this specific symbol (check for `%automagic` in https://ipython.readthedocs.io/en/stable/interactive/magics.html). This log error basically says that **maybe** the program failed on the file because of that magic command conversion. As far as I know, those magic commands **probably** won't make our program fail, because we are only looking for imports with the AST tree.

Fernando-crz:
Just fixed installing jupyter notebooks as an optional dependency. Basically, the program detects if the user has the library `nbconvert` installed, and if it is installed, the program will start detecting jupyter notebooks automatically when generating the _requirements.txt_ file.

I was thinking of maybe turning the detection of jupyter notebooks when creating the _requirements.txt_ file optional, using an optional argument with the name "`--ignore-notebooks`". What do you think? @alan-barzilay

pipreqs/pipreqs.py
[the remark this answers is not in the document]
alan-barzilay: @Fernando-crz then why are we passing None at all? why not always default to the python standar that is utf-8? I think a new commit to change this pipreqs wide could be good idea

pipreqs/pipreqs.py
alan-barzilay: @Fernando-crz first thing would be to understand why is it doing this haha are we sure this happens in the latest version of black? what are the exact options set? can anyone replicate this weird formatting?

pipreqs/pipreqs.py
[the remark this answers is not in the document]
alan-barzilay: please add a test to check if magic commands without % would really make pipreqs to fail. If it is not an issue we can drop this if

pipreqs/pipreqs.py
alan-barzilay: 23.9.1^ output of black version with flag for 120 character lines

setup.py
alan-barzilay: please add instructions to read me about how to deal with these optional dependencies

alan-barzilay: also, if possible please reverse the behaviour so that by default the jupyter stuff is added and it is only excluded as a dependency when explicitly told to

alan-barzilay:
> I was thinking of maybe turning the detection of jupyter notebooks when creating the _requirements.txt_ file optional, using an optional argument with the name "`--ignore-notebooks`". What do you think? @alan-barzilay

yeah, good idea!

alan-barzilay (changes requested):
great work, we are almost there!

pipreqs/pipreqs.py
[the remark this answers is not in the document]
Fernando-crz: Done

pipreqs/pipreqs.py
```diff
@@ -488,6 +488,9 @@ def init(args):
     extra_ignore_dirs = args.get("--ignore")
     follow_links = not args.get("--no-follow-links")
     input_path = args["<path>"]
+
```
alan-barzilay, line 491: @Fernando-crz  woof woof

Fernando-crz:
> I was thinking of maybe turning the detection of jupyter notebooks when creating the _requirements.txt_ file optional, using an optional argument with the name "`--ignore-notebooks`". What do you think? @alan-barzilay

Done.

pipreqs/pipreqs.py
Fernando-crz: Done

pipreqs/pipreqs.py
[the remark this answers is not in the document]
Fernando-crz: Just created a new file for magic commands inside the jupyter notebook test case (the file is named `magic_commands.ipynb`). As it turns out, tests are running successfully. It seems that magic commands with or without the percentage sign ("_%_") don't affect the ast search. 

With that in mind, i deleted the **if** related to this issue.

pipreqs/pipreqs.py
Fernando-crz: I suggest that we can use darker instead (https://github.com/akaihola/darker/tree/master). It only runs the formatter over changes in the commits, meaning that, if I change this formatting back to it's origin state, it won't be changed in the next commits to the "next" branch.

By now, I will just revert the changes related to this specific line.

pipreqs/pipreqs.py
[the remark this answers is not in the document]
Fernando-crz: Oddly enough, just noticed my tests ran successfully when run individually but failed when running with tox. I'll fix this asap

pipreqs/pipreqs.py
```diff
@@ -145,11 +178,35 @@ def get_all_imports(path, encoding=None, extra_ignore_dirs=None, follow_links=Tr
     return list(packages - data)
```
mateuslatrova, line 179:
```suggestion

def get_file_extensions():
    return [".py", ".ipynb"] if PythonExporter and not ignore_notebooks else [".py"]
```

pipreqs/pipreqs.py
```diff
@@ -102,13 +122,26 @@ def get_all_imports(path, encoding=None, extra_ignore_dirs=None, follow_links=Tr
         dirs[:] = [d for d in dirs if d not in ignore_dirs]
 
         candidates.append(os.path.basename(root))
-        files = [fn for fn in files if os.path.splitext(fn)[1] == ".py"]
+        if PythonExporter and not ignore_notebooks:
+            files = [fn for fn in files if filter_ext(fn, [".py", ".ipynb"])]
+        else:
+            files = [fn for fn in files if filter_ext(fn, [".py"])]
```
mateuslatrova, line 128:
```suggestion
            extensions =  get_file_extensions()
            files = [fn for fn in files if filter_ext(fn, extensions)]
```

pipreqs/pipreqs.py
```diff
@@ -145,11 +178,35 @@ def get_all_imports(path, encoding=None, extra_ignore_dirs=None, follow_links=Tr
     return list(packages - data)
 
 
+def filter_ext(file_name, acceptable):
```
mateuslatrova, line 181: guess we could improve this functions name to something more meaningful.

something like `def file_extension_is_allowed(file_name, allow_list)`

pipreqs/pipreqs.py
```diff
@@ -102,13 +122,26 @@ def get_all_imports(path, encoding=None, extra_ignore_dirs=None, follow_links=Tr
...
-        candidates += [os.path.splitext(fn)[0] for fn in files]
         for file_name in files:
             file_name = os.path.join(root, file_name)
-            with open(file_name, "r", encoding=encoding) as f:
-                contents = f.read()
+            contents = ""
+            if filter_ext(file_name, [".py"]):
+                with open(file_name, "r", encoding=encoding) as f:
+                    contents = f.read()
+            elif filter_ext(file_name, [".ipynb"]) and PythonExporter and not ignore_notebooks:
```
mateuslatrova, line 143: not the first time I see this condition `PythonExporter and not ignore_notebooks`.

maybe we could create a function with this condition so that its meaning is clearer. Example:

```
def notebooks_are_enabled():
   return PythonExporter and not ignore_notebooks
```

tests/test_pipreqs.py
```diff
@@ -517,6 +519,48 @@ def test_output_requirements(self):
...
+        for item in imports:
+            self.assertTrue(item.lower() in self.modules, "Import is missing: " + item)
+        self.assertFalse("time" in imports)
+        self.assertFalse("logging" in imports)
+        self.assertFalse("curses" in imports)
+        self.assertFalse("__future__" in imports)
+        self.assertFalse("django" in imports)
+        self.assertFalse("models" in imports)
+        self.assertFalse("FastAPI" in imports)
+        self.assertFalse("sklearn" in imports)
```
mateuslatrova, line 537:
```suggestion
        not_desired_imports = ["time", "logging", "curses", "__future__", "django", "models", "FastAPI", "sklearn"]        
        self.assertTrue(all(import not in imports for import in not_desired_imports))
```

tests/test_pipreqs.py
```diff
@@ -559,6 +603,28 @@ def test_parse_requirements_handles_file_not_found(self, exit_mock):
...
+                "--pypi-server": None,
+                "--print": False,
+                "--diff": None,
+                "--clean": None,
+                "--mode": None,
+                "--ignore-notebooks": True,
+            }
+        )
+        assert os.path.exists(self.requirements_notebook_path) == 1
+        assert os.path.getsize(self.requirements_notebook_path) <= 1
```
mateuslatrova, line 626: what does this check exactly?

pipreqs/pipreqs.py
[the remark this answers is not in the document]
Fernando-crz: Just fixed the issue. As it turns out, tox was not considering the jupyter optional dependency when running it's tests, thus making my local test diverge from tox tests. Just added a line in the `tox.ini` doc to consider the optional dependency when testing. @alan-barzilay

setup.py
Fernando-crz: > also, if possible please reverse the behaviour so that by default the jupyter stuff is added and it is only excluded as a dependency when explicitly told to

It seems like there's no support for doing such thing. Check https://stackoverflow.com/questions/76376575/python-setuptools-exclude-dependencies-when-installing

What do you suggest doing in this case? Drop optional support? Maybe keep it as it is?

Fernando-crz: > What do you suggest doing in this case? Drop optional support? Maybe keep it as it is?

@alan-barzilay

tests/test_pipreqs.py
Fernando-crz: Done

pipreqs/pipreqs.py
Fernando-crz: Done

pipreqs/pipreqs.py
Fernando-crz: Changed it to `file_ext_is_allowed` to avoid having too large names for functions

pipreqs/pipreqs.py
Fernando-crz: Decided to not implement this function because it makes the file extensions used by the program "hidden". The user will only know the file extension being used in the program if the user reads the function directly, making our code a little bit less readable

pipreqs/pipreqs.py
Fernando-crz: > Decided to not implement this function because it makes the file extensions used by the program "hidden". The user will only know the file extension being used in the program if the user reads the function directly, making our code a little bit less readable

^^

tests/test_pipreqs.py
Fernando-crz: It just checks if the `requirements.txt` file generated by this test is created and is empty.

setup.py
alan-barzilay: hi, sorry for the delay. If there is no support for this feature, there is no support for it, what can we do haha
But i liked that `pip install --no-deps my_package` workaround, if we only import nbconverter/ipython when we use them (aka when we don't pass the cli option to ignore notebooks), someone that does this should be able to tailor their one dependencies without too much trouble. Maybe we should add a mention to this into the readme as well

alan-barzilay: and, maybe I missed it but where do we use ipython directly? I found the import for nbvconvert, is it that ipython is a dependency of nbconvert? if thats the case, cant we omit it from the file and let nbconvert install its dependencies itself?

tests/test_pipreqs.py
```diff
@@ -517,6 +515,42 @@ def test_output_requirements(self):
...
+
+    def test_ipynb_2_py(self):
+        """
+        Test the function ipynb_2_py() which converts .ipynb file to .py format
+        """
+        expected = pipreqs.get_all_imports(self.compatible_files["original"])
+        parsed = pipreqs.get_all_imports(self.compatible_files["notebook"])
+        self.assertEqual(expected, parsed)
+
+        parsed = pipreqs.get_all_imports(self.compatible_files["notebook"], encoding="utf-8")
```
alan-barzilay, line 543: if the encoding is already utf-8, why are you adding it here? or is it from the original PR? if so, please update it accordingly

pipreqs/pipreqs.py
```diff
@@ -48,9 +49,19 @@
 from yarg import json2package
 from yarg.exceptions import HTTPError
 
+try:
+    PythonExporter = None
+    ignore_notebooks = False
+    from nbconvert import PythonExporter
+except ImportError:
+    pass
+
```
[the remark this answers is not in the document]
alan-barzilay: oh, was the idea for it to be inside the exception handling? or are you making the existence of Python Exporter take precedence before the ignore notebooks flag?

alan-barzilay (changes requested):
good work as always, dont forget to start thinking about how to adapt this PR to support poetry

.github/workflows/tests.yml
```diff
@@ -39,7 +39,8 @@ jobs:
             - name: Install dependencies
               run: |
                   python -m pip install --upgrade pip
-                  pip install coverage docopt yarg requests
+                  pip install coverage docopt yarg requests nbconvert
+                  pip install .[jupyter]
```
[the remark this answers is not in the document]
Fernando-crz: "`.[jupyter]`" means that we are installing a library located in the current working directory and we are installing the optional depencies specified by the "`jupyter`" version. In this case, it means that we are installing _pipreqs_ with support for jupyter notebooks.

pipreqs/pipreqs.py
```diff
@@ -145,11 +178,39 @@ def get_all_imports(path, encoding=None, extra_ignore_dirs=None, follow_links=Tr
     return list(packages - data)
 
 
+def notebooks_are_enabled():
+   return PythonExporter and not ignore_notebooks
```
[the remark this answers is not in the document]
Fernando-crz: woofn't

pipreqs/pipreqs.py
```diff
@@ -256,7 +317,7 @@ def get_import_local(imports, encoding=None):
     # had to use second method instead of the previous one,
     # because we have a list in the 'exports' field
     # https://stackoverflow.com/questions/9427163/remove-duplicate-dict-in-list-in-python
-    result_unique = [i for n, i in enumerate(result) if i not in result[n + 1:]]
+    result_unique = [i for n, i in enumerate(result) if i not in result[n + 1 :]]
```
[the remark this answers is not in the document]
Fernando-crz: done

setup.py
alan-barzilay: please test if we really need the ipython dependency for  magics translation

tests/test_pipreqs.py
alan-barzilay: make this its one commit

alan-barzilay:
Just so we have a written record of what we've just discussed:  let's drop the idea of optional dependencies.   Since we want the default option to include support for jupiter notebook, having optional dependencies don't  bring us any value.

 Besides that we also discussed the idea to further abstract some of the code processing steps to make them more auto contained and hide some of the logic with abstractions.  We'll also hide the jupiter logic behind the global variable ignore notebooks, if it is set to true nbconverter will never be imported.

@Fernando-crz  did i forget to mention anything?

alan-barzilay:
@Fernando-crz can we close this PR if it is being continued in #414 ?

Fernando-crz:
> Just so we have a written record of what we've just discussed: let's drop the idea of optional dependencies. Since we want the default option to include support for jupiter notebook, having optional dependencies don't bring us any value.
> 
> Besides that we also discussed the idea to further abstract some of the code processing steps to make them more auto contained and hide some of the logic with abstractions. We'll also hide the jupiter logic behind the global variable ignore notebooks, if it is set to true nbconverter will never be imported.
> 
> @Fernando-crz did i forget to mention anything?

LGMT

Fernando-crz:
> @Fernando-crz can we close this PR if it is being continued in #414 ?

Sounds good to me.

PR #408, opened 2023-10-20
add workflow for automatic deploy on pypi

mateuslatrova:
With this workflow, whenever we create a new release on GitHub, GitHub Actions will deploy that new release to PyPi


.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install build
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
```
alan-barzilay, line 36: @mateuslatrova  how does it know the correct package name? is it associated with the token? Have you tried uploading it to some dummy test package? PYPI has a test "version" for this sort of thing

mateuslatrova: @alan-barzilay the package name is gotten from the "`name`" parameter in the "`setup`" function call inside "`setup.py`" file, which in this case is _pipreqs_.

yes, i have tested it in my PyPi account. I ran this pipeline for my fork of the pipreqs repo, and it worked.

if you want to, we can schedule a call for me to show you how it worked, and also what you will need to do for this to work in production

.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,14 @@
...
+on:
+  release:
+    types: [published]
+jobs:
+  build_and_publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build and publish to pypi
+        uses: JRubics/poetry-publish@v1.17
```
alan-barzilay, line 11: what made you choose this action specifically instead of other poetry github actions? like: 
https://github.com/marketplace/actions/python-poetry-action
or
https://github.com/marketplace/actions/install-poetry-action

daherati: I think, `JRubics/poetry-publish@v1.17`, used for building and publishing python packages to pypi, pypi-test or a private wheels repo using poetry. and the two you mentioned here mostly focus on actions installing and configuring poetry.

alan-barzilay:
@mateuslatrova can we close this PR if its being worked on #411?

PR #415, opened 2023-11-08
Test badge

alan-barzilay:
[update status badge from travis to github action](https://github.com/bndr/pipreqs/commit/e2c0f6483df4a447adc42079d66617adee35ed4c) and [Break tests and codecov workflow in 2](https://github.com/bndr/pipreqs/commit/eabd8b91ff044c708d25f4936288fc2a0798d63f).

The idea is to have the status badge report if the tests are passing,
without considering codecov since it frequently fails due to some
network hickup that makes the report upload to fail.

This also makes things a bit more compartimentalized.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,30 @@
+name: CodeCov
+on:
+    workflow_run:
+        workflows: [Tests]
+        types: [completed]
+
+jobs:
+    coverage_report:
+        runs-on: ubuntu-latest
+        if: ${{ github.event.workflow_run.conclusion == 'success' }}
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v3
+
+            - name: Install dependencies
+              run: |
+                  python -m pip install --upgrade pip
+                  pip install coverage docopt yarg requests
+
+            - name: Calculate coverage
+              run: coverage run --source=pipreqs -m unittest discover
+
+            - name: Create XML report
+              run: coverage xml
+
+            - name: Upload coverage to Codecov
+              uses: codecov/codecov-action@v3
+              with:
+                  files: coverage.xml
+                  fail_ci_if_error: true
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,4 +1,4 @@
-name: Tests and Codecov
+name: Tests
 on:
   push:
   pull_request:
@@ -28,27 +28,3 @@ jobs:
 
             - name: Test with tox
               run: tox
-
-    coverage_report:
-        needs: run_tests
-        runs-on: ubuntu-latest
-        steps:
-            - name: Checkout repository
-              uses: actions/checkout@v3
-
-            - name: Install dependencies
-              run: |
-                  python -m pip install --upgrade pip
-                  pip install coverage docopt yarg requests
-
-            - name: Calculate coverage
-              run: coverage run --source=pipreqs -m unittest discover
-
-            - name: Create XML report
-              run: coverage xml
-
-            - name: Upload coverage to Codecov
-              uses: codecov/codecov-action@v3
-              with:
-                  files: coverage.xml
-                  fail_ci_if_error: true
--- a/README.rst
+++ b/README.rst
@@ -2,9 +2,8 @@
 ``pipreqs`` - Generate requirements.txt file for any project based on imports
 =============================================================================
 
-.. image:: https://img.shields.io/travis/bndr/pipreqs.svg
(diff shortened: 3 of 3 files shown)


alan-barzilay:
relevant reference: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run


Also:

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

alan-barzilay:
it will look something like this, but without the codecov part since I broke the workflow down into 2
image

alan-barzilay:
tested on https://github.com/pipreqsxp/pipreqs/actions/runs/6800189760/job/18488179557 and tests work as expected once they reach the main branch, with codecov being called after a successful run of "Tests"

README.rst
```diff
@@ -2,9 +2,8 @@
 ``pipreqs`` - Generate requirements.txt file for any project based on imports
 =============================================================================
 
-.. image:: https://img.shields.io/travis/bndr/pipreqs.svg
-        :target: https://travis-ci.org/bndr/pipreqs
-
+.. image:: https://github.com/bndr/pipreqs/actions/workflows/tests.yml/badge.svg?branch=master
```
alan-barzilay, line 5: I set the tracking to tests performed on the master branch, a broken test in the next branch or in a random PR doesnt seem to be very informative or relevant to users since unless they manually instal pipreqs they will get the master branch through pypi

.github/workflows/codecov.yml
```diff
@@ -0,0 +1,30 @@
...
+              run: |
+                  python -m pip install --upgrade pip
+                  pip install coverage docopt yarg requests
+
+            - name: Calculate coverage
+              run: coverage run --source=pipreqs -m unittest discover
+
+            - name: Create XML report
+              run: coverage xml
+
```
willianrocha, line 25: Would you consider adding a report here? like it is done in [Code Coverage Report](https://github.com/marketplace/actions/code-coverage-report-action) using insightsengineering/coverage-action@v2. It would be nice to see the status during the PR progress.

.github/workflows/codecov.yml
```diff
@@ -0,0 +1,30 @@
+name: CodeCov
+on:
+    workflow_run:
+        workflows: [Tests]
+        types: [completed]
+
+jobs:
+    coverage_report:
+        runs-on: ubuntu-latest
+        if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
willianrocha, line 10: If the pipeline fails, will the badge reflect update the status?

.github/workflows/codecov.yml
```diff
@@ -0,0 +1,30 @@
...
+              run: coverage run --source=pipreqs -m unittest discover
+
+            - name: Create XML report
+              run: coverage xml
+
+            - name: Upload coverage to Codecov
+              uses: codecov/codecov-action@v3
+              with:
+                  files: coverage.xml
+                  fail_ci_if_error: true
```
willianrocha, line 30: consider adding an empty line at the file end

.github/workflows/tests.yml
```diff
@@ -28,27 +28,3 @@ jobs:
 
             - name: Test with tox
               run: tox
```
willianrocha, line 30: consider adding an empty line at the file end

.github/workflows/codecov.yml
alan-barzilay: nice! I wanted something like that but at the time I wrote the task I hadnt found anything of this sort, will try to add some sort of report

alan-barzilay:
> Would you consider adding a report here? like it is done in [Code Coverage Report](https://github.com/marketplace/actions/code-coverage-report-action) using insightsengineering/coverage-action@v2. It would be nice to see the status during the PR progress.


That would be nice but sadly we run into a problem of not having permission for the bot to do its thing. In the interest of time I will drop this idea for now and create an issue to register our interest for the future.

Here is the step I tried to implement into the workflow:
```
            - name: Produce the coverage report
              uses: insightsengineering/coverage-action@v2
              with:
		path: coverage.xml
		# Publish the rendered output as a PR comment
		publish: true
		# Create a coverage diff report.
		diff: true
		# Branch to diff against.
		diff-branch: next
		# This is where the coverage reports for the
		# `diff-branch` are stored.
		diff-storage: _xml_coverage_reports
		togglable-report: true
```

alan-barzilay:
there seems to be an issue, the codecov test is run but doesnt show up in the PR, it seems to loose any association with what trigerred it

.github/workflows/codecov.yml
alan-barzilay: moved to issue #417

PR #421, opened 2023-11-27
Implementing darker as pre-commit

willianrocha:
Implementing pre-commit rules to apply linting only in codes changed.

It was implemented a set of rules, using [pre-commit](https://pre-commit.com/), to run linting before commit is made.

For linting tool, kept using [flake8](https://flake8.pycqa.org/en/latest/), but added [black](https://github.com/python/black).
On top of these two tools, [Darker](https://github.com/akaihola/darker) was used to apply the linting rules only to changed lines.


willianrocha:
In flynt 0.68, some breaking changes were introduced as states in darker release notes [1.7.1](https://github.com/akaihola/darker/blob/master/CHANGES.rst#171---2023-03-26)

mateuslatrova:
Hey @willianrocha!

1. There are some merge conflicts, so I guess this branch needs rebasing.

2. It seems to me that most of the changes were related to formatting (correct me if I'm wrong). Since the important changes don't include many code/configuration changes and not all commits have a working version of the feature, maybe you could reduce the number of commits to one or two(which would be working versions), and describe any important information(like the breaking changes from flynt) in the commits' description.

3. When I run `git commit -m 'test'`, I receive the following error: `FileNotFoundError: [Errno 2] No such file or directory: 'flake8'`. Then I ran the same command with poetry: `poetry run git commit -m 'test'` and it worked. Is this the expected behavior? It does not seem so intuitive. But if it stays like this, I guess it should be documented in the `CONTRIBUTING.rst` file.

CONTRIBUTING.rst
```diff
@@ -67,6 +67,10 @@ Ready to contribute? Here's how to set up `pipreqs` for local development.
 
     $ poetry install --with dev
 
+3.1. Pipreqs is using pre-commits
+
+    $ poetry run pre-commit install
```
alan-barzilay, line 72: can' t this be done in the `poetry install --with dev` step somehow?

alan-barzilay: since pre commit is already marked as a dependency, do we need this? or is it installing the hooks and not  package pre-commit?

.pre-commit-config.yaml
```diff
@@ -13,6 +13,5 @@ repos:
       - id: darker
         args: [--isort, --skip-string-normalization]
         additional_dependencies:
-          - black~=23.11.0
-          - isort~=5.9
-          - flynt~=0.78
+          - isort==5.9
```
alan-barzilay, line 16: why this versions exactly?

pyproject.toml
```diff
@@ -57,10 +56,12 @@ line-length = 120
...
 [tool.black]
-exclude = [
-    "tests/_data/",
-    "tests/_data_clean/",
-    "tests/_data_duplicated_deps/",
-    "tests/_data_ignore/",
-    "tests/_invalid_data/",
-]
+exclude = '''
+/(
```
alan-barzilay, line 60: whats the difference between this and the old format? it was more readable/pythonic, was it broken?

alan-barzilay (changes requested):
I make matheus words my own, good work but the commit messages need deeper explanations and justifications for the choices made. I agree everything could be one or 2 commits, maybe one for the new files and the second one with the changes made by applying the pre-commit.

Since there is that issue with flynt and blacker, I assume there is no better alternative, right? I guess we gotta live with that weird version exigence haha

alan-barzilay:
> When I run git commit -m 'test', I receive the following error: FileNotFoundError: [Errno 2] No such file or directory: 'flake8'. Then I ran the same command with poetry: poetry run git commit -m 'test' and it worked. Is this the expected behavior? It does not seem so intuitive. But if it stays like this, I guess it should be documented in the CONTRIBUTING.rst file.

forgot to emphasize this one, git commit failing is not acceptable behaviour haha

Issue #214, opened 2020-07-16
UnicodeDecodeError: 'utf-8' codec can't decode byte


tbrodbeck:
I just installed pipreqs for the first time.
This is what it does:
```
$ pipreqs $(pwd)
Traceback (most recent call last):
  File "/Users/tillmann/dev/textSum/.direnv/python-3.8.3/bin/pipreqs", line 10, in <module>
    sys.exit(main())
  File "/Users/tillmann/dev/textSum/.direnv/python-3.8.3/lib/python3.8/site-packages/pipreqs/pipreqs.py", line 470, in main
    init(args)
  File "/Users/tillmann/dev/textSum/.direnv/python-3.8.3/lib/python3.8/site-packages/pipreqs/pipreqs.py", line 406, in init
    candidates = get_all_imports(input_path,
  File "/Users/tillmann/dev/textSum/.direnv/python-3.8.3/lib/python3.8/site-packages/pipreqs/pipreqs.py", line 122, in get_all_imports
    contents = f.read()
  File "/usr/local/bin/../Cellar/python@3.8/3.8.3_2/bin/../Frameworks/Python.framework/Versions/3.8/lib/python3.8/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb1 in position 81: invalid start byte
```
Am I doing something wrong?

```
$ neofetch           
                    'c.          tillmann@MacBook-Pro 
                 ,xNMM.          -------------------- 
               .OMMMMo           OS: macOS Catalina 10.15.5 19F101 x86_64 
               OMMM0,            Host: MacBookPro15,1 
     .;loddo:' loolloddol;.      Kernel: 19.5.0 
   cKMMMMMMMMMMNWMMMMMMMMMM0:    Uptime: 17 days, 14 hours, 5 mins 
 .KMMMMMMMMMMMMMMMMMMMMMMMWd.    Packages: 115 (brew) 
 XMMMMMMMMMMMMMMMMMMMMMMMX.      Shell: zsh 5.7.1 
;MMMMMMMMMMMMMMMMMMMMMMMM:       Resolution: 1680x1050@2x, 3840x1600@2x, 1050x1680@2x 
:MMMMMMMMMMMMMMMMMMMMMMMM:       DE: Aqua 
.MMMMMMMMMMMMMMMMMMMMMMMMX.      WM: Quartz Compositor 
[...]

PR #439, opened 2024-03-14
Added debug message to describe decode failures discussed in issue 214

KyleLuoma:
I added this exception handler to deal with #214 which, although closed, can still cause grief if the user is not able to ignore sub directories that contain the culprit file. This does not resolve the decode error; but it provides an additional debug statement that names the file causing the error.

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -177,8 +177,11 @@ def get_file_extensions():
 
 def read_file_content(file_name: str, encoding="utf-8"):
     if file_ext_is_allowed(file_name, DEFAULT_EXTENSIONS):
-        with open(file_name, "r", encoding=encoding) as f:
-            contents = f.read()
+        try:
+                contents = f.read()
+        except Exception as e:
+            logging.debug("Encountered exception when attempting to decode: {0}".format(file_name))
+            raise e
     elif file_ext_is_allowed(file_name, [".ipynb"]) and scan_noteboooks:
         contents = ipynb_2_py(file_name, encoding=encoding)
     return contents


kmulka-bloomberg:
This looks like an improvement over the current situation. I wonder if files like these should just be ignorned and the code moves on to the next one.

StantonMatt (commented):
I checked this locally because #214 still has a long tail of users trying to identify the file that triggered the decode error. The goal makes sense, but the current diff has a blocking regression: `read_file_content()` no longer opens the file before calling `f.read()`, so ordinary `.py` files fail before the decode path is reached.

Minimal repro on this branch:

```text
.venv/bin/python - <<'PY'
from pipreqs.pipreqs import read_file_content
read_file_content("tests/_data/test.py")
PY
```

That raises:

```text
NameError: name 'f' is not defined
```

The same direct probe on current `origin/master` returns the file contents normally.

Other local checks I ran:

```text
.venv/bin/python -m py_compile pipreqs/pipreqs.py tests/test_pipreqs.py
.venv/bin/python -m flake8 pipreqs tests
.venv/bin/python -m unittest discover
```

`py_compile` and `git diff --check` passed. `flake8` fails on `pipreqs/pipreqs.py:181` with `E117` and `F821 undefined name 'f'`. The unit suite fails 18 tests, all through the same `read_file_content()` `NameError`. GitHub does not report any check runs for this PR.

I think this needs the `with open(file_name, "r", encoding=encoding) as f:` block restored inside the `try`, plus a regression test that covers both a normal Python file read and the decode-error logging path. If the intended behavior is to skip unreadable files instead of re-raising, that should probably be a separate explicit behavior choice because it would change how #214 is handled.

PR #440, opened 2024-03-19
Add mapping `jose` -> `python-jose`

Pwuts:
`jose` is a Python 2 package. `python-jose` is its Python 3 successor.


jonas-eschle:
Good point, thanks. That being said, maybe you could convince the python-jose developers to contact pypi and take over the Jose name? With the last release 10 years ago, this seems like an inactive repo...
LGTM!

PR #445, opened 2024-04-26
Add support for scanning imports in specified virtual environments, fix #434

### LINKED ISSUE #434: pipreqs reporting same module with two versions
I ran pipreqs for the first time, and after fixing some syntax errors it found in my code, it spat out a shorter list than I expected (haven't confirmed anything is missing yet), but what really caught my attention was it repeated two modules with two versions.

Output:
`GitPython==3.1.40
GitPython==3.1.41
jsonpickle==3.0.2
numpy==1.26.4
openpyxl==3.1.2
pandas==2.2.0
platformdirs==4.2.0
PyQt5==5.15.10
PyQt5_sip==12.13.0
PySide6==6.6.0
PySide6==6.6.1
PySide6_Addons==6.6.0
PySide6_Essentials==6.6.0
Requests==2.31.0`

As you can see, GitPython and PySide6 are both repeated with slightly differing versions.

Output from the terminal:
`PS C:\FolderName> pipreqs --force
WARNING: Import named "GitPython" not found locally. Trying to resolve it at the PyPI server.
WARNING: Import named "GitPython" was resolved to "GitPython:3.1.41" package (https://pypi.org/project/GitPython/).
Please, verify manually the final list of requirements.txt to avoid possible dependency confusions.
WARNING: Import named "numpy" not found locally. Trying to resolve it at the PyPI server.
WARNING: Import named "numpy" was resolved to "numpy:1.26.4" package (https://pypi.org/project/numpy/).
Please, verify manually the final list of requirements.txt to avoid possible dependency confusions.
WARNING: Import named "pandas" not found locally. Trying to resolve it at the PyPI server.
WARNING: Import named "pandas" was resolved to "pandas:2.2.0" package (https://pypi.org/project/pandas/).
Please, verify manually the final list of requirements.txt to avoid possible dependency confusions.
WARNING: Import named "platformdirs" not found locally. Trying to resolve it at the PyPI server.
WARNING: Import named "platformdirs" was resolved to "platformdirs:4.2.0" package (https://pypi.org/project/platformdirs/).
Please, verify manually the final list of requirements.txt to avoid possible dependency confusions.
WARNING: Import named "PyQt5" not found locally. Trying to resolve it at the PyPI server.
WARNING: Import named "PyQt5" was resolved to "PyQt5:5.15.10" package (https://pypi.org/project/PyQt5/).
Please, verify manually the final list of requirements.txt to avoid possible dependency confusions.
WARNING: Import named "PySide6" not found locally. Trying to resolve it at the PyPI server.
WARNING: Import named "PySide6" was resolved to "PySide6:6.6.1" package (https://pypi.org/project/PySide6/).
Please, verify manually the final list of requirements.txt to avoid possible dependency confusions.
WARNING: Import named "Requests" not found locally. Trying to resolve it at the PyPI server.
WARNING: Import named "Requests" was resolved to "requests:2.31.0" package (https://pypi.org/project/requests/).
Please, verify manually the final list of requirements.txt to avoid possible dependency confusions.
INFO: Successfully saved requirements file in C:\FolderName\requirements.txt`

While it does say "Please, verify manually the final list of requirements.txt to avoid possible dependency confusions." this is repeated for every version it found, so this is not a useful warning. These warnings seem more like normal operation too, not a warning.

Tdarnell:
Added a argument "--venv" which looks for imports in only specified folders, defaults to sys.path when this is not given

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,7 +25,7 @@ Installation
 
     pip install pipreqs
 
-Obs.: if you don't want support for jupyter notebooks, you can install pipreqs without the dependencies that give support to it. 
+Obs.: if you don't want support for jupyter notebooks, you can install pipreqs without the dependencies that give support to it.
 To do so, run:
 
 .. code-block:: sh
@@ -66,6 +66,7 @@ Usage
                               <gt>     | e.g. Flask>=1.1.2
                               <no-pin> | e.g. Flask
         --scan-notebooks      Look for imports in jupyter notebook files.
+        --venv <dirs>...      Look for imports in the specified virtualenv
 
 Example
 -------
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -36,6 +36,7 @@ Options:
                           <gt>     | e.g. Flask>=1.1.2
                           <no-pin> | e.g. Flask
     --scan-notebooks      Look for imports in jupyter notebook files.
+    --venv <dirs>...      Look for imports in the specified virtualenv
 """
 from contextlib import contextmanager
 import os
@@ -109,7 +110,9 @@ def get_all_imports(path, encoding="utf-8", extra_ignore_dirs=None, follow_links
         ".tox",
         "__pycache__",
         "env",
+        ".env",
         "venv",
+        ".venv",
         ".ipynb_checkpoints",
     ]
 
@@ -259,10 +262,10 @@ def get_imports_info(imports, pypi_server="https://pypi.python.org/pypi/", proxy
     return result
 
 
-def get_locally_installed_packages(encoding="utf-8"):
+def get_locally_installed_packages(encoding="utf-8", paths: list = sys.path):
     packages = []
     ignore = ["tests", "_tests", "egg", "EGG", "info"]
-    for path in sys.path:
+    for path in paths:
         for root, dirs, files in os.walk(path):
             for item in files:
                 if "top_level" in item:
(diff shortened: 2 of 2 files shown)


Tdarnell:
This is in relation to issue #434

jonas-eschle:
Hi @Tdarnell , thanks a lot for the PR. I understand the issue, however, I am not sure if this is the best way to go about it. I agree that the version should match the installed version if somehow possible, but this rather fixes a special case. What about conda envs for example?

StantonMatt (commented):
I checked this against #434 locally. The path-filtering approach does solve the narrow duplicate-version case when the caller points pipreqs at the environment it should trust: with a fixture source importing a package that only existed in the supplied metadata directory, `--venv <path> --print` emitted the version from that directory. Passing a comma-separated pair of metadata roots also worked.

What I ran on this branch:

```sh
python -m unittest discover
```

All passed. GitHub reports no checks for this PR branch.

One behavior gap: the help text says `--venv <dirs>...`, but the implementation actually expects a single comma-separated value. `pipreqs <src> --venv <dir1> <dir2> --print` fails with the usage message, while `--venv <dir1>,<dir2>` works. So before merging, I’d either change the option/help text to describe comma-separated paths, or adjust the parsing to support the documented repeated-argument form. A small regression test for the #434 path-selection case would also make this easier to evaluate without recreating the fixture manually.

PR #467, opened 2024-11-29
Avoid test/tests folder to ignore the file encoding issue, beasuse so…

EircYangQM:
Avoid test/tests folder to ignore the file encoding issue, beasuse some package like IPython, joblib use the other encoding file for testing.


jonas-eschle:
Hm, I think this isn't just an encoding issue but rather a breaking change as the test/tests folder will  be ignored, right?

PR #499, opened 2025-09-01
add file input error handling

MYusufY:
This PR adds error handling for the case when the user provides a file instead of a directory.

# Example:

## Current version
`pipreqs project/main.py`:
```
Traceback (most recent call last):
  File "/usr/local/bin/pipreqs", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pipreqs/pipreqs.py", line 609, in main
    init(args)
  File "/usr/local/lib/python3.12/site-packages/pipreqs/pipreqs.py", line 599, in init
    generate_requirements_file(path, imports, symbol)
  File "/usr/local/lib/python3.12/site-packages/pipreqs/pipreqs.py", line 209, in generate_requirements_file
    with _open(path, "w") as out_file:
  File "/usr/local/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/pipreqs/pipreqs.py", line 91, in _open
    file = open(filename, mode)
           ^^^^^^^^^^^^^^^^^^^^
NotADirectoryError: [Errno 20] Not a directory: 'project/main.py/requirements.txt'
```
## This fork
`pipreqs project/main.py`:
`ERROR: The path 'project/main.py' is not a directory. Please provide a directory path.
`

# Testing

This works properly in my tests.

### Question

Is this section in contributing guideline optional?:

> To run a subset of tests:
> 
> $ poetry run python -m unittest tests.test_pipreqs

Thanks!

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -516,6 +516,9 @@ def init(args):
         encoding = "utf-8"
     if input_path is None:
         input_path = os.path.abspath(os.curdir)
+    if not os.path.isdir(input_path):
+        logging.error(f"The path '{input_path}' is not a directory. Please provide a directory path.")
+        sys.exit(1)
 
     if extra_ignore_dirs:
         extra_ignore_dirs = extra_ignore_dirs.split(",")


alan-barzilay (commented):
its been a while since I handled this codebase, but pretty sure pipreqs can be run with an individual file as a target so this would throw an error in a legitimate use case.

Also, the project is a bit abandoned but we previously had a policy of requiring tests for any changes and new features.  I dont think there are new maintainers but even if there are they may not follow this policy anymore. Do with this information what you will, sorry I cant be a bigger help

StantonMatt (commented):
I checked this locally because the current bare-file target failure is real, but Alan's concern about valid file targets also reproduces.

Validation on the PR branch:

- `python -m py_compile pipreqs/pipreqs.py tests/test_pipreqs.py`
- `python -m flake8 pipreqs tests`
- `git diff --check origin/master...HEAD`
- `python -m unittest discover`

No upstream checks are reported on this PR branch.

On current `origin/master`, a file target works when the caller does not ask pipreqs to create `file.py/requirements.txt`:

- `pipreqs tests/_data/test.py --print` exits 0
- `pipreqs tests/_data/test.py --savepath .codex-tmp/pipreqs-499-master-file-target.txt --force` exits 0

The broken case is the bare file target without `--print` or `--savepath`, which still reproduces the reported `NotADirectoryError` by trying to write `tests/_data/test.py/requirements.txt`.

With this PR, all three forms exit 1 before scanning:

- `pipreqs tests/_data/test.py --print`
- `pipreqs tests/_data/test.py --savepath ... --force`
- `pipreqs tests/_data/test.py --force`

So I think the guard is too broad. A safer fix would keep file targets valid when output goes to stdout or an explicit save path, and only reject or handle the no-output-path case that would otherwise build `file.py/requirements.txt`. A regression test for both forms would be useful here:

- file target + `--print` or `--savepath` should still work
- file target without `--print`/`--savepath` should get the friendly error instead of `NotADirectoryError`

PR #528, opened 2026-04-10
Fix local package detection for packages without top_level.txt

### LINKED ISSUE #374: pipreqs says "WARNING: Import named "..." not found locally. Trying to resolve it at the PyPI server."
Hi all,

This is on MacOS Catalina. My project uses python 3.10, but I also have python 3.8 installed as well as python 2.7 as it is used by the system.

I run

pipreqs .

in the project directory and for every import it finds in the code issues

"WARNING: Import named "..." not found locally. Trying to resolve it at the PyPI server."

even though these are in fact installed for python 3.10.

How do I make it find them?

Thanks.

datatalking:
Adds fallback detection for .dist-info directories that lack top_level.txt (e.g., scipy, seaborn). Parses package name and version directly from the directory name when top_level.txt is missing.

Fixes #374


StantonMatt (commented):
I checked this locally because #374 has a long trail of reports around packages like scipy/seaborn missing `top_level.txt`, and the fallback does cover that core shape.

With a synthetic `scipy-1.9.3.dist-info` directory and no `top_level.txt`:

```text
origin/master packages []
origin/master import local []
pr-528 packages [{'name': 'scipy', 'version': '1.9.3', 'exports': ['scipy']}]
pr-528 import local [{'name': 'scipy', 'version': '1.9.3', 'exports': ['scipy']}]
```

So the intended local-detection path works in that direct repro. The branch does need a bit more test work before I would call it ready, though. Full local discovery currently fails:

```text
python -m unittest discover
FAILED (failures=1)
```

The failing test is `tests.test_pipreqs.TestPipreqs.test_init_local_only`. In my environment, `--use-local` now writes:

```text
beautifulsoup4==4.14.3
docopt==0.6.2
ipython==8.18.1
pyflakes==3.4.0
```

The new `beautifulsoup4` entry comes from the existing `bs4` import mapping plus the new fallback seeing `beautifulsoup4-4.14.3.dist-info` without `top_level.txt`. That may be intended behavior, but the test expectation has not been updated, and there is not yet a committed regression test for the scipy/seaborn-style no-`top_level.txt` case.

Other checks passed:

```text
python -m flake8 pipreqs/pipreqs.py tests/test_pipreqs.py
```

GitHub reports no check runs for this PR. I think the next step is to add a small fixture/regression test for a `.dist-info` directory without `top_level.txt`, and either update `test_init_local_only` for the now-detected `beautifulsoup4` output or adjust the fallback if that output is not desired.

hernot:
@datatalking, @StantonMatt:  Just some thoughts - I guess you thought your self about and had good reasons not to go this route.
According [Python Packaging Guide](https://packaging.python.org/en/latest/specifications/recording-installed-packages/) METADATA is the only mandatory file in `*.dist-info` diretories. It has to contain package name and version. 
So how about the following METADATA centered approach in replacement for simple "fallbacking"?

1) parse the METADATA file in all `*.dist-info` directories for **name** and **version** fields and add both to list of local pacages
2) check if `top_level.txt` is present in `*.dist-info` directory and add its content to package list as done now

**Edit**
-------------
I checked **pyserial** - a package i know is to be imported as `serial` instead of **pyserial** -> it has a `top_level.txt` file in its `.dist-info` directory. Whereas eg. `numpy` and `scipy`, where package and import names match, don't have. With out rechecking and reading in detail the above mentioned Packaging guide is guess the `top_level.txt` only needs to be present when import and package name differ or when multiple sub-packages and modules should be importable as if distinct top-level modules and packages.

bubaizhanshen:
I gave this another local pass after @hernot's METADATA note.

The core no-`top_level.txt` case does work: a synthetic `scipy-1.9.3.dist-info` with only `METADATA` is detected as local for `import scipy`.

I still do not think the current fallback is ready as-is, for two reasons:

1. `test_init_local_only` still fails for me on Python 3.10. With the current branch it writes:

```text
beautifulsoup4==4.14.3
docopt==0.6.2
ipython==8.39.0
pyflakes==3.4.0
```

That may be the intended new behavior, but the test expectation has not been updated.

2. The fallback ignores `METADATA`'s canonical `Name` and uses the dist-info directory stem instead. For example, with:

```text
my_package-1.2.3.dist-info/METADATA
Name: my-package
Version: 1.2.3
```

this branch records `name='my_package'`, not `my-package`. Pip normalization may make that installable, but reading `Name`/`Version` from METADATA would avoid guessing and would match the packaging spec better.

The basic checks pass locally:

```text
python -m flake8 pipreqs/pipreqs.py tests/test_pipreqs.py
```

PR #529, opened 2026-05-25
Fixes #58 Implement egg-info version parsing

### LINKED ISSUE #58: <package>/<package>.egg-info directories cause bad versions in requirements.txt
I usually install multiple packages under development in a virtualenv using `python setup.py develop`. As a result, I get a `<package>.egg-info` directory (which contains `top_level.txt` etc.).

In presence of such packages, the `requirements.txt` file created by `pipreqs` contains entries like this:

    mypackage.egg==info
    otherpackage.egg==info

Instead, either the version from `setup.py` should be used, or maybe the version should be omitted altogether.

(see also #33 which was about checking local versions of packages)

arnowaschk:
Fixes #58

  `pipreqs` could misparse `<package>.egg-info` metadata directories and emit malformed package/version pairs. This patch:

  - prefers `METADATA` / `PKG-INFO` when available
  - preserves the old version fallback for regular directory names
  - adds regression coverage for egg-info and metadata-based parsing

  I verified the focused regression tests locally.

Greetings Arno


StantonMatt (commented):
I checked this locally against #58. The branch fixes the original no-version `<package>.egg-info` case in a way that keeps the package name usable instead of splitting the suffix as a bogus version.

Direct reproduction:

```text
origin/master: [{'name': 'mypackage.egg', 'version': 'info', 'exports': ['mypackage']}]
pr-529:       [{'name': 'mypackage', 'version': None, 'exports': ['mypackage']}]
```

I also checked the metadata path and the old directory-name fallback through the new tests. Local verification on Python 3.9.6:

```text
python -m unittest tests.test_pipreqs.TestPipreqs.test_get_locally_installed_packages_ignores_egg_info_suffix tests.test_pipreqs.TestPipreqs.test_get_package_name_and_version_preserves_regular_version tests.test_pipreqs.TestPipreqs.test_get_package_name_and_version_keeps_hyphenated_name_without_version tests.test_pipreqs.TestPipreqs.test_get_locally_installed_packages_prefers_metadata_name
python -m unittest discover
python -m flake8 pipreqs/pipreqs.py tests/test_pipreqs.py
```

The focused tests passed, full discovery passed 33 tests, and lint/compile/diff checks passed. GitHub reports no check runs for this PR, so there is no upstream CI result to claim, but I do not see a blocking issue from this pass.

arnowaschk:
@StantonMatt Thank you for your efforts!

PNHD (approved):
Reviewed current head `cbe034f`. The helper now prefers standards-based METADATA/PKG-INFO name/version data and fixes the `.egg-info` suffix fallback without regressing ordinary versioned directory names. The added regressions cover the original no-version egg-info case, metadata preference, and the legacy directory-name fallback; an independent local full-suite/lint pass is also recorded on the PR. LGTM.

Issue #360, opened 2023-03-01
Pipreqs adds wrong package to requirements.txt


artemgur:
In my project, I use kafka-python package.
When pipreqs generates requirements.txt, it adds kafka package in addition to kafka-python package.
kafka package is not used in the project, it's not even installed.

Arfius:
I have the same problem.

p4r4xor:
I suppose pipreqs doesn't support hyphenated packages yet.

Godd67:
I have similar issue with azure-storage.
It adds azure-storage==0.37, which is deprecated and cannot be installed instead of azure-storage-blob==12.15.0, that is actually installed and used

CwjXFH:
It looks like the same problem as [sklearn not correctly formed](https://github.com/bndr/pipreqs/issues/358)

kevindugan:
> It adds azure-storage==0.37, which is deprecated and cannot be installed instead of azure-storage-blob==12.15.0, that is actually installed and used. Is there a workaround for this or should I switch back to pip freeze to create requirements files?

I'm having the same issue with azure related modules. It seems like there are a few assumptions in the package deduction that are getting in the way.
1. For azure specifically, there are multiple entries in the mapping with the 'azure' key. Meaning that subsequent entries for 'azure' will overwrite previous entries. Hence, why `azure-storage` is showing up.
2. The `get_all_imports` function assumes that we only want the top-level module. They give the example `django.conf` should only import `django`. This assumption is broken for several of the azure modules, eg `azure.functions` actually should point to the azure-functions project.

Currently I don't have a great proposal for how to redesign the mapping between module imports and pypi project names. I think this is a recurring issue that the maintainers have identified. My current thought is there may be a way to programatically generate the mapping file by queries to the pypi package api, though I don't have a working prototype yet.

LockeKK:
I had the same problem.
I have a private python module locates in folder lib, then pipreqs added lib into requirement.txt while this lib acctully refers to https://pypi.org/project/lib/ and confused users that install with this txt file.

bobfirestone:
> I had the same problem.
> I have a private python module locates in folder lib, then pipreqs added lib into requirement.txt while this lib acctully refers to https://pypi.org/project/lib/ and confused users that install with this txt file.

I am building an internal proof of concept python monorepo tool and get the same thing but with `pkg` instead of `lib`.

[...]

Issue #141, opened 2018-11-09
Resolve import name collisions intelligently


jonafato:
In some cases (e.g. #138), an import name can point to multiple packages and result in outputting a suboptimal or incorrect requirement. `pipreqs` could make better decisions here using some heuristics such as existing requirements files, currently installed libraries, or external popularity / usage statistics from PyPI or GitHub. This would also require the mapping file to support one-to-many mappings.

PR #531, opened 2026-06-01
Map common import names to their package names

### LINKED ISSUE #491: Issue with libraries with hyphens "-" in the name
I have a project with "cv2" and "scikit-learn" but when i used the command
`pipreqs . --force --mode no-pin`
It was giving the requirements.txt as 

> noise
> numpy
> opencv_python
> Pillow
> scikit_learn
> scipy
> skimage
> torch
> torchvision
> tqdm

As you can see the opencv-python is written as opencv_python (which is in correct manner "opencv-python") and more of the bad one is "skimage" which is really "scikit-image".

If I am doing something wrong please guide me.

bubaizhanshen:
Fixes #491. Addresses part of #360.

This maps common import names to their package names for `kafka`, `janitor`, `skimage`, and `sklearn`, and adds regression coverage for those aliases.

Checks:
- `python -m unittest tests.test_pipreqs.TestPipreqs.test_get_pkg_names tests.test_pipreqs.TestPipreqs.test_get_pkg_names_maps_common_import_aliases`
- `python -m py_compile pipreqs/pipreqs.py tests/test_pipreqs.py`
- `flake8 tests/test_pipreqs.py`
- `git diff --check`


StantonMatt (commented):
I checked this branch locally. The mapping changes are consistent with the existing import-to-distribution override path, and the new test covers the four aliases directly.

Validation from a project-local Python 3.9 venv:

- `PYTHONPATH=. python -m unittest tests.test_pipreqs.TestPipreqs.test_get_pkg_names tests.test_pipreqs.TestPipreqs.test_get_pkg_names_maps_common_import_aliases` -> 2 tests, OK
- `PYTHONPATH=. python -m py_compile pipreqs/pipreqs.py tests/test_pipreqs.py` -> OK
- `python -m flake8 tests/test_pipreqs.py` -> OK
- `git diff --check origin/master...HEAD` -> OK

Only local note: the unittest command emitted urllib3's LibreSSL warning from this macOS Python; it was not from this patch.

PNHD (changes requested):
There is one blocking collision in the new mapping set: `janitor:pyjanitor` is not an unambiguous import-to-distribution mapping. PyPI also has a distinct `janitor` distribution, while `pyjanitor` exposes the same top-level `janitor` import. Hard-mapping that import to `pyjanitor` can therefore rewrite a legitimate dependency to a different project. This is the class of one-import-to-multiple-distributions problem already tracked in #141. Please remove/split the ambiguous `janitor` entry (or resolve it through a collision-aware mechanism); the `kafka`, `skimage`, and `sklearn` fixes can then be evaluated independently.

Issue #469, opened 2025-01-06
Unicode Decode Error


hadi-f90:
### Problem

 In Python 3.12 running pipreqs returns this error:
```
<unknown>:165: SyntaxWarning: invalid escape sequence '\S'
<unknown>:166: SyntaxWarning: invalid escape sequence '\['
<unknown>:207: SyntaxWarning: invalid escape sequence '\['
<unknown>:456: SyntaxWarning: invalid escape sequence '\S'
<unknown>:200: SyntaxWarning: invalid escape sequence '\:'
Traceback (most recent call last):
  File "/*******/.local/bin/pipreqs", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/*******/.local/lib/python3.12/site-packages/pipreqs/pipreqs.py", line 609, in main
    init(args)
  File "/*******/.local/lib/python3.12/site-packages/pipreqs/pipreqs.py", line 533, in init
    candidates = get_all_imports(
                 ^^^^^^^^^^^^^^^^
  File "/*******/.local/lib/python3.12/site-packages/pipreqs/pipreqs.py", line 136, in get_all_imports
    contents = read_file_content(file_name, encoding)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/*******/.local/lib/python3.12/site-packages/pipreqs/pipreqs.py", line 181, in read_file_content
    contents = f.read()
               ^^^^^^^^
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb1 in position 81: invalid start byte
```
Therefore, it doesn't save any requirements.txt

### Possible Solution
It seems that changing the afformentioned piece of code into this one can manage to solve it temporarily:

```
        with open(file_name, "r", encoding=encoding) as f:
            try:
                contents = f.read()
            except UnicodeDecodeError:
                contents = f.read().enode("utf-8")
```

Although changing this solves the main issue about saving requirements.txt file, it adds raising of a number of warnings about the name and need to check module/lib names from PyPi which slows down pipreqs.

py563:
+1, same issue on Windows with python 3.13.1 on 0.4 versions of pipreqs too

janekolszak:
Nope, this doesn't work

JvdsReform:
I have the same problem

mzglinski:
Try ignoring the venv folder if you have one, for example

```
[...]

PR #532, opened 2026-06-01
Expand default ignored directories

### LINKED ISSUE #420: Ignore dotfiles directories by default
I was hitting major issues because of the .venv and .mypy directories in my project. 

- The common issue with virtual environments is [already covered in this issue](https://github.com/bndr/pipreqs/issues/214), but ignoring dot-prefixed directories would fix this for 80% of people.
- Mypy stores cache files in directories named by the package being analyzed, so pipreqs thinks they are local packages and does not register them as external dependencies.

I can't think of a reason why source files should be in a dot directory, but if that's not a surefire rule, an alternative could be to copy [the default ignored paths that `isort` uses](https://pycqa.github.io/isort/docs/configuration/options.html#skip) ([source](https://github.com/PyCQA/isort/blob/main/isort/settings.py#L72-L94)):
```
DEFAULT_SKIP: FrozenSet[str] = frozenset(
    {
        ".venv",
        "venv",
        ".tox",
        ".eggs",
        ".git",
        ".hg",
        ".mypy_cache",
        ".nox",
        ".svn",
        ".bzr",
        "_build",
        "buck-out",
        "build",
        "dist",
        ".pants.d",
        ".direnv",
        "node_modules",
        "__pypackages__",
        ".pytype",
    }
)
```

I'd suggest taking a page from their book on [recognizing modules types](https://github.com/PyCQA/isort/blob/main/isort/place.py#L15) as well. Just pass in the module name, and ignore anything that's not "THIRDPARTY". It's not 100%, but it's tried and mostly true.

StantonMatt:
## Summary
- centralize the default scanner ignore list for reuse
- ignore common generated/cache/environment directories such as .mypy_cache, .nox, build, dist, node_modules, and __pypackages__ by default
- add coverage that files inside default ignored directories do not contribute imports

Fixes #420.

## Verification
- .venv/bin/python -m unittest tests.test_pipreqs.TestPipreqs.test_get_all_imports_skips_default_ignored_directories
- .venv/bin/python -m unittest tests.test_pipreqs
- .venv/bin/python -m flake8 pipreqs tests
- review-fix-loop: clean fresh review, no actionable correctness issues


bubaizhanshen:

The new default-ignore path works for the issue shape from #469/#495: with `main.py` importing `requests` and a `.venv/lib/bad.py` file containing undecodable bytes plus `import click`, `get_all_imports()` returns only `['requests']`. So the scanner no longer reads into the virtualenv by default.

Checks I ran:

```text
python -m unittest tests.test_pipreqs.TestPipreqs.test_get_all_imports_skips_default_ignored_directories tests.test_pipreqs.TestPipreqs.test_get_all_imports
```

All passed. I did not use `test_ignored_directory` as a signal here because it waits on a live PyPI request in this environment, but the default directory-pruning behavior changed by this PR looks good from this pass.

PNHD (approved):
Reviewed current head `5ee4ffb` against #420. Centralizing the default ignore set and adding common generated/cache/environment directories prevents scanner noise without changing the explicit user ignore path, and the regression verifies imports under those defaults do not leak into results. This is a focused implementation of the issue's requested default-ignore behavior. LGTM.

PR #534, opened 2026-06-01
Honor Python source encoding declarations

### LINKED ISSUE #461: Support for files that specify their encoding
I ran into the same issue many had here:
https://github.com/bndr/pipreqs/issues/214

Python source code does have a standard way of specifying the encoding of a file. I suggest pipreqs uses this if available to detect the encoding of the file.

PEP-263: Defining Python Source Code Encodings
https://peps.python.org/pep-0263/

StantonMatt:
## Summary

- use Python's `tokenize.open()` for normal `.py`/`.pyw` scanning so PEP 263 source encoding declarations are honored
- preserve explicit non-default `--encoding` behavior
- add a regression for a cp1252-encoded Python file containing a non-UTF-8 byte

Fixes #461.

## Verification

- `.venv/bin/python -m unittest tests.test_pipreqs.TestPipreqs.test_python_source_encoding_cookie`
- `uv run --with flake8 flake8 pipreqs tests`
- `git diff --check`
- `review-fix-loop`: clean


bubaizhanshen:

The PEP 263 path works as intended: a `# coding: cp1252` source file containing a non-UTF-8 byte is scanned with the default encoding path and still returns `['requests']`. I also checked that explicit non-default encoding still works for a cp1252 file without a source cookie.

Checks I ran:

```text
python -m unittest tests.test_pipreqs.TestPipreqs.test_python_source_encoding_cookie tests.test_pipreqs.TestPipreqs.test_get_all_imports
```

All passed.

PNHD (approved):
Reviewed current head `dd79c49d`. Using `tokenize.open()` on the default Python-source path is the standard-library mechanism for honoring PEP 263 encoding declarations and UTF-8 BOMs, while explicit non-default `--encoding` still takes the existing direct-open path. The cp1252 regression exercises the reported failure mode, and the change is narrowly scoped to Python source reads. LGTM.

PR #536, opened 2026-06-01
Protect custom savepath from accidental overwrite

### LINKED ISSUE #336: pipreqs --savepath always overwrite despite i'm not using --force
Hi,

i'm actually using pipreqs to generate a requirements.in file and then use pip-tools on it.

My issue is that using `pipreqs --savepath requirements.in` is overwriting my existing requirements.in file, even if i do not use the --force option.

Any idea why this is happening? Is there a way to just append the new requirements pipreqs finds without overwriting the existing ones? (Some i added by hand, so they get deleted everytime).

This does not happen when using pipreqs to generate the "default" requirements.txt, so the issue is only when specifying a custom file name with --savepath.

Thank you.

StantonMatt:
## Summary

- apply the existing overwrite protection to the resolved output path, including custom `--savepath` files
- keep `--print`, `--diff`, `--clean`, and stdout output from being blocked by the file-generation guard
- add regression coverage for custom savepath no-force, force, diff, and clean behavior

Fixes #336.

## Verification

- `.venv/bin/python -m unittest tests.test_pipreqs.TestPipreqs.test_init_savepath tests.test_pipreqs.TestPipreqs.test_init_overwrite tests.test_pipreqs.TestPipreqs.test_init_savepath_does_not_overwrite_existing_file_without_force tests.test_pipreqs.TestPipreqs.test_init_savepath_force_overwrites_existing_file tests.test_pipreqs.TestPipreqs.test_init_savepath_existing_file_does_not_block_diff tests.test_pipreqs.TestPipreqs.test_init_savepath_existing_file_does_not_block_clean`
- `uv run --with flake8 flake8 pipreqs tests`
- `git diff --check`
- `review-fix-loop clean`


bubaizhanshen:

The output-path guard behaves correctly in a direct probe:

- existing custom `--savepath` without `--force` returns before scanning and keeps the file unchanged
- `--force` overwrites the existing custom savepath
- existing custom `--savepath` does not block `--diff` or `--clean`

The focused unit command from the PR body waits on a live PyPI request in this environment through the older `test_init_savepath`, so I did not use that as the main signal. The behavior above was checked with `get_all_imports`, `get_import_local`, `diff`, and `clean` mocked where appropriate.

Static checks passed:

```text
```

PNHD (approved):
Reviewed current head `19204bf`. The overwrite guard now applies to the resolved output path, so an existing custom `--savepath` is protected unless `--force` is set. The guard correctly stays out of `--diff`, `--clean`, `--print`, and stdout paths, and those modes return before file generation. The added regressions cover the custom-savepath protection and the mode exemptions. LGTM.

Issue #90, opened 2017-10-09
Print to stdout by default


timwis:
Thanks for this tool! I just wanted to suggest that perhaps `pipreqs` print to stdout by default, allowing you to redirect to a file via:

```bash
pipreqs > requirements.txt
```

But this would let you check the results before writing (or overwriting) to the file, by seeing it in your terminal first. Another option is to use a `-o requirements.txt` flag.

Just a suggestion :)

jonafato:
While this isn't the default, it is possible using the `--print` flag. When using this flag, the output is written to stdout rather than a file. If you'd like to change the filename without printing to stdout, you can do so with the `--savepath` flag.

The "standard" way to provide these options in other tools are `--output filename` and `--output -` (for stdout). Adding these as synonyms for the existing options may make this behavior more conventionally discoverable.

jonafato:
@timwis do the existing options address your use case?

timwis:
Sure, it works just fine -- I just wanted to suggest flipping the default to be more similar to other unix CLI tools. But just an opinion/suggestion -- no right or wrong answer. I'll close for now :)

jonafato:
I think this makes sense for a future release (one that would include breaking changes). Just wanted to make sure you knew these options existed today.

KKZ20:
Thanks for this tool very much! But I wonder if there can be a option such as --no-info-output to cancel the output information "Successfully saved requirements file in xxx"? Because I want to use the pipreqs instruction by subprocess in my python project but I just don't want to output this info in my shell. Just a suggestion.

TomerGodinger:
I'd like to second what KKZ20 said here (both the message of gratitude and the remark about the info output).
It's a bit strange in my opinion for an info message specifying success to be printed to `stderr` in the first place, but I suppose it's better than mixing it in with the output.
In general it would be nice if this tool acted more similarly to standard UNIX tools, printing the results to `stdout` by default, printing to `stderr` only when an error occurs and just printing nothing if it's successful but there are no dependencies to print.
The indication of success seems rather redundant to me. It should be conveyed through the exit code instead.

Well, at least this can be dealt with using the currently available options.
Once again, thanks for making this. =)

PR #539, opened 2026-06-02
Add --output option alias

StantonMatt:
## Summary

- Add `--output <file>` as a conventional alias for the existing `--savepath` behavior.
- Add `--output -` as a conventional alias for the existing `--print` stdout behavior.
- Cover both aliases with focused `init()` tests.

This keeps the default behavior unchanged while addressing the discoverability suggestion from #90.

## Tests

- `.venv/bin/python -m unittest tests.test_pipreqs.TestPipreqs.test_init_output_path tests.test_pipreqs.TestPipreqs.test_init_output_dash`
- `.venv/bin/flake8 pipreqs tests`
- `git diff --check`
- `review-fix-loop`: no actionable correctness issues


bubaizhanshen:

The alias behavior works in a direct probe with local package lookup mocked:

- `--output out.txt` writes the generated requirements to that file
- `--output -` writes to stdout and does not create the default `requirements.txt`

Static checks passed:

```text
```

The focused unit command from the PR body waits on a live PyPI request in this environment through `test_init_output_path`, so I did not use it as the main signal here.

PNHD (approved):
Reviewed current head `060bde0` against #90. `--output <file>` delegates to the existing save-path behavior and `--output -` delegates to the existing stdout path, so the alias does not introduce a second output implementation. Both forms have focused `init()` coverage and the existing default behavior remains unchanged. LGTM.

PR #545, opened 2026-07-07
Add popular utility package mappings (attrs, python-docx, XlsxWriter)

Johnnnmai:
Adds mappings for widely-used utility packages. For example, `import attr` requires `attrs` on PyPI, and `import docx` requires `python-docx`.

### DIFF
--- a/pipreqs/mapping
+++ b/pipreqs/mapping
@@ -320,6 +320,7 @@ atreal:atreal.richfile.preview
 atreal:atreal.richfile.qualifier
 atreal:atreal.usersinout
 atsim:atsim.potentials
+attr:attrs
 attractsdk:attract_sdk
 audio:audio.bitstream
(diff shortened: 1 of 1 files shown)


PNHD (changes requested):
Two of these one-to-one mappings are ambiguous and can produce incorrect requirements. `attr` is itself a distinct PyPI distribution that provides `from attr import attr` (and explicitly notes that `attrs` also provides/shadows an `attr` package), while both `whois` and `python-whois` are separate distributions using the `whois` import name. Hard-mapping `attr -> attrs` and `whois -> python-whois` therefore rewrites valid dependencies to different projects. #141 already tracks this one-import-to-multiple-distributions problem. Please remove/split the ambiguous entries unless the resolver can disambiguate them; the unambiguous mappings can be reviewed separately.

PR #547, opened 2026-07-07
Add system/platform package mappings (wxPython, PyGObject, setuptools)

Johnnnmai:
Adds mappings for system-level packages. Notably, `import pkg_resources` comes from setuptools, and `import gi` comes from PyGObject.


PNHD (changes requested):
`pkg_resources:setuptools` is no longer a safe current mapping. `pkg_resources` was removed from setuptools in v82.0.0, so generating an unpinned `setuptools` requirement can install a current setuptools release that still does not provide the imported module. Please remove/split this entry, or only represent it if the mapping/output model can encode the required historical version constraint. The `Xlib`, `gi`, and `wx` mappings do not need to be blocked on that change.

PR #542, opened 2026-07-07
Add explicit UTF-8 encoding to internal file reads

Johnnnmai:
## Summary

Three `open()` calls in `pipreqs/pipreqs.py` read files without specifying an encoding, relying instead on the platform's default encoding (`locale.getpreferredencoding()`). On systems where that default isn't UTF-8 — for example GBK on Chinese Windows, or Latin-1 on some European locales — these reads fail with `UnicodeDecodeError`.

This affects:
- Fixes #241
- Fixes #271
- Fixes #469

## Changes

- `open(join("stdlib"), "r")` -> add `encoding="utf-8"` (reads pipreqs' own stdlib module list)
- `open(join("mapping"), "r")` -> add `encoding="utf-8"` (reads pipreqs' own import-to-package mapping file)
- `open(file_, "r")` in `parse_requirements` -> add `encoding="utf-8"`

The `stdlib` and `mapping` files are pipreqs' own bundled data files and are always UTF-8, so pinning the encoding explicitly is safe and makes behavior consistent across all platforms/locales instead of depending on the user's system default encoding.

## Test plan

- [x] Manually inspected all three call sites and confirmed `encoding="utf-8"` is a safe, no-behavior-change fix on UTF-8 systems
- [ ] Reproduce on a non-UTF-8-default system (e.g. Chinese Windows with GBK locale) and confirm `pipreqs` no longer raises `UnicodeDecodeError`

Issue #271, opened 2021-09-03
Error if script contains Umlauts (like ä,ö,ü)


sgaechter:
pipreqs trows Error when script contains umlauts:
Stack trace:
```
ERROR: Failed on file: P:\PROJECTS\Dataplot_Preislistendownload\Dataplot_Preisliste.py
Traceback (most recent call last):
  File "c:\python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Python39\Scripts\pipreqs.exe\__main__.py", line 7, in <module>
  File "c:\python39\lib\site-packages\pipreqs\pipreqs.py", line 470, in main
    init(args)
  File "c:\python39\lib\site-packages\pipreqs\pipreqs.py", line 406, in init
    candidates = get_all_imports(input_path,
  File "c:\python39\lib\site-packages\pipreqs\pipreqs.py", line 138, in get_all_imports
    raise exc
  File "c:\python39\lib\site-packages\pipreqs\pipreqs.py", line 124, in get_all_imports
    tree = ast.parse(contents)
  File "c:\python39\lib\ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 89
    df = df.assign(WiederverkÃ¤ufer = round_col(vk - (vk /WVfaktor),2))
                              ^
```
attached affecting Script (please change to *.py :-) )
[Dataplot_Preislist.txt](https://github.com/bndr/pipreqs/files/7104757/Dataplot_Preislist.txt)

sgaechter:
Hello @hasansezertasan probably it's a OS bug. From your post, I gather that you reproduced it with Linux. I use Windows. I will try to narrow it down even further.

Issue #241, opened 2021-04-06
UnicodeDecodeError: 'gbk' codec can't decode byte 0x8c in position 358: illegal multibyte sequence


wellitecho:
# What happened
I used pipreqs to freeze package requirements for one of my projects by running 'pipreqs' in the project directory, but it caused this error: UnicodeDecodeError: 'gbk' codec can't decode byte 0x8c in position 358: illegal multibyte sequence.

# What I expected
It should work without causing any problem.

# Full traceback
Traceback (most recent call last):
  File "c:\users\myAccount\anaconda3\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\myAccount\anaconda3\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\myAccount\anaconda3\Scripts\pipreqs.exe\__main__.py", line 7, in <module>
  File "c:\users\myAccount\anaconda3\lib\site-packages\pipreqs\pipreqs.py", line 470, in main
    init(args)
  File "c:\users\myAccount\anaconda3\lib\site-packages\pipreqs\pipreqs.py", line 406, in init
    candidates = get_all_imports(input_path,
  File "c:\users\myAccount\anaconda3\lib\site-packages\pipreqs\pipreqs.py", line 122, in get_all_imports
    contents = f.read()
UnicodeDecodeError: 'gbk' codec can't decode byte 0x8c in position 358: illegal multibyte sequence

# What I did then
I went to c:\users\myAccount\anaconda3\lib\site-packages\pipreqs\pipreqs.py and modified lines near line 122(contents = f.read()), namely, the lines concerning file encoding:
original:
```python
                with open_func(file_name, "r", encoding=encoding) as f:
                    contents = f.read()

```
after adding try/except:
```python
            try:
                with open_func(file_name, "r", encoding=encoding) as f:
                    contents = f.read()
            except:
                with open_func(file_name, "r", encoding='utf-8') as f:
                    contents = f.read()
```
Then it worked without problems. Albeit a temp fix.

# What need to be done
The assumption about the encoding of the source code file need to be revised?

wellitecho:
#214 
Maybe we should dynamically decide what the encoding of the file is?

alan-barzilay:
I think revisiting the encoding option is a good idea as well, I think I will make utf-8 the default since (if I'm not mistaking) it is the default encoding for python3.
[...]

PR #548, opened 2026-07-07
Fix encoding in _open() and clean() functions

Johnnnmai:
Continuation of #542. Adds explicit UTF-8 encoding to two remaining open() calls:
- `_open()` context manager (line ~92): used when writing requirements.txt
- `clean()` function (line ~459): used when cleaning requirements.txt

Without these, pipreqs can produce garbled output or crash on systems with non-UTF-8 default encoding.

Fixes part of #241, #271, #469.

### DIFF
--- a/pipreqs/pipreqs.py
+++ b/pipreqs/pipreqs.py
@@ -89,7 +89,7 @@ def _open(filename=None, mode="r"):
         else:
             raise ValueError("Invalid mode for file: {}".format(mode))
     else:
-        file = open(filename, mode)
+        file = open(filename, mode, encoding="utf-8") if "b" not in mode else open(filename, mode)
 
     try:
         yield file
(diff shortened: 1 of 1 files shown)


PNHD (changes requested):
The new-output path and the in-place clean path need different encoding treatment.

"
        "Using an explicit encoding when creating a new requirements file is reasonable, but `clean(file_, imports)` opens an existing user file with `r+`. Hard-coding UTF-8 there changes current behavior and can make an existing non-UTF8 requirements file unreadable on a system where it currently works with the locale default. Since `clean` rewrites an existing file, it should preserve/detect that file's encoding (or accept the configured encoding) and have a regression covering a non-UTF8 existing requirements file.

"
        "The issue references are also mismatched: #241, #271, and #469 fail while scanning Python source, whereas this diff only changes the output helper and `clean()`. Please do not claim those source-scanning issues are fixed by this patch.

PR #550, opened 2026-07-07
ci: Add Windows and macOS to test matrix

Johnnnmai:
The existing CI only runs on Ubuntu, but several open bugs (#241, #271, #469) are platform-specific encoding issues that only manifest on Windows (GBK default encoding) and European Linux/macOS (Latin-1 default).

Adding Windows and macOS to the test matrix ensures platform-specific regressions are caught. This is particularly important given the recent encoding fixes in #542 and #548.

Changes:
- Added `windows-latest` and `macos-latest` to the OS matrix
- Excluded Python 3.9 on macOS (unavailable on ARM runners)
- Removed pypy to keep CI time manageable

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,11 +10,15 @@ on:
 
 jobs:
     run_tests:
-        runs-on: ubuntu-24.04
+        runs-on: ${{ matrix.os }}
         strategy:
             fail-fast: false
             matrix:
-                python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.10']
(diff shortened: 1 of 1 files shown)


PNHD (changes requested):
Adding Windows/macOS coverage is useful, but this patch also removes the existing PyPy coverage. `tox.ini` explicitly includes `pypy3` in `envlist` and maps the GitHub Actions `pypy-3.10` interpreter to that environment, so dropping it is an unrelated reduction in implementation coverage rather than a necessary part of the OS expansion. Please retain PyPy on Ubuntu (for example via a matrix `include` entry or a separate job) while adding the CPython Windows/macOS matrix.

